PR #11, opened 2023-01-16
Bugfix/cplp 2036 offer submit failing on missing salesmanager

ntruchsess:
implements https://jira.catena-x.net/browse/CPLP-2036


ntruchsess:
incorrect target repo

PR #20, opened 2023-04-13
fix(notification): create only one notification per user

Phil91:
the notifications will only be created once per request for each user and type

Refs: CPLP-2526


Phil91:
Closed because of wrong target

PR #33, opened 2023-05-16
feat(app): add endpoint to change subscription url

Phil91:
## Description

The following endpoint has been added to change the subscription url for subscriptions that are active: 

/api/apps/appchanges/{appId}/subscription/{subscriptionId}/tenantUrl

## Why

To enable a provider to have the possibilty to change the instance url for an offer for a specific subscription

## Issue

N/A - Jira Ticket: CPLP-2656

## Checklist


src/marketplace/Apps.Service/BusinessLogic/AppChangeBusinessLogic.cs
```diff
@@ -197,4 +199,77 @@ public async Task UploadOfferAssignedAppLeadImageDocumentByIdAsync(Guid appId, s
...
+        {
+            throw new ConflictException($"There is no subscription detail data configured for subscription {subscriptionId}");
+        }
+
+        if (url == detailData.SubscriptionUrl)
+            return;
+        
+        offerSubscriptionsRepository.AttachAndModifyAppSubscriptionDetail(detailData.DetailId, subscriptionId, os =>
+        {
+            os.AppSubscriptionUrl = url;
```
ntruchsess, line 245: AppSubscriptionUrl is nullable - to allow deletion of SubscriptionUrl the property must be initialized to the previous value before attaching the entity to the context

tests/marketplace/Apps.Service.Tests/BusinessLogic/AppChangeBusinessLogicTest.cs
```diff
@@ -599,4 +595,304 @@ public async Task DeactivateOfferStatusbyAppIdAsync_CallsExpected()
...
+            {
+                setParameters.Invoke(details);
+            });
+
+        // Act
+        await _sut.UpdateTenantUrlAsync(appId, subscriptionId, data, _iamUserId).ConfigureAwait(false);
+
+        // Assert
+        A.CallTo(() => _provisioningManager.UpdateClient(clientClientId, data.Url, A<string>._)).MustHaveHappenedOnceExactly();
+        A.CallTo(() => _notificationService.CreateNotifications(A<IDictionary<string, IEnumerable<string>>>._, A<Guid?>._, A<IEnumerable<(string? content, NotificationTypeId notificationTypeId)>>._, A<Guid>._, A<bool?>._)).MustHaveHappenedOnceExactly();
```
ntruchsess, line 680: It also needs to be checked the all the results are actually being awaited, otherwise no notification would be sent out. (use Org.Eclipse.TractusX.Portal.Backend.Tests.Shared.Extensions.FakeIAsyncEnumerableExtensions.AsFakeIAsyncEnumerable(out var asyncEnumerator) to get access to the IAsyncEnumerator that you need to assert MoveNextAsync being called the expected number of times).

tests/portalbackend/PortalBackend.DBAccess.Tests/OfferSubscriptionRepositoryTest.cs
```diff
@@ -293,6 +293,83 @@ public async Task GetSubscriptionDetailForProviderAsync_WithWrongUser_ReturnsExp
...
+            });
+
+        // Assert
+        var changeTracker = context.ChangeTracker;
+        var changedEntries = changeTracker.Entries().ToList();
+        changeTracker.HasChanges().Should().BeTrue();
+        changedEntries.Should().NotBeEmpty();
+        changedEntries.Should().HaveCount(1);
+        changedEntries.Single().Entity.Should().BeOfType<AppSubscriptionDetail>()
+            .Which.AppSubscriptionUrl.Should().Be(modifiedUrl);
```
ntruchsess, line 368: need to verify this works for modifiedUrl being null as well.
I did adjust the unit-test which makes it fail as expected. Please pull and adjust the repository-method accordingly to allow setting modificationUrl to null.

ntruchsess (changes requested):

tests/portalbackend/PortalBackend.DBAccess.Tests/OfferSubscriptionRepositoryTest.cs
Phil91: added a check to test whether the update would be executed when the url doesn't change as well

tests/marketplace/Apps.Service.Tests/BusinessLogic/AppChangeBusinessLogicTest.cs
Phil91: added the check

src/marketplace/Apps.Service/BusinessLogic/AppChangeBusinessLogic.cs
Phil91: good point, adjusted the repository and bl code and added a unit test for it.

ntruchsess (approved):

PR #67, opened 2023-06-07
feat(connectors):- Delete Connector

qxz2mqe:
## Description
Delete connector for scenarios like when daps,DS and statusid flag are false, true and active pending etc

## Why

For connector registrations where the daps registration was unsuccessful, the deletion is currently blocked.

## Issue
https://jira.catena-x.net/browse/CPLP-2780

## Checklist

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
@@ -289,49 +289,50 @@ public class ConnectorsBusinessLogic : IConnectorsBusinessLogic
     public async Task DeleteConnectorAsync(Guid connectorId, Guid userId, CancellationToken cancellationToken)
     {
         var connectorsRepository = _portalRepositories.GetInstance<IConnectorsRepository>();
-        var result = await connectorsRepository.GetConnectorDeleteDataAsync(connectorId).ConfigureAwait(false);
-        var (IsConnectorIdExist, DapsClientId, SelfDescriptionDocumentid, DocumentStatus, ConnectorStatus, DapsRegistrationSuccess) = result;
-        if (!IsConnectorIdExist)
+        var (isValidConnectorId, dapsClientId, selfDescriptionDocumentId, 
+        documentStatus, connectorStatus, dapsRegistrationSuccess) = await 
+        connectorsRepository.GetConnectorDeleteDataAsync(connectorId).ConfigureAwait(false);
+
+        if (!isValidConnectorId)
         {
             throw new NotFoundException($"Connector {connectorId} does not exist");
         }
-        if ((SelfDescriptionDocumentid == null) && DapsRegistrationSuccess!.Value && ConnectorStatus == ConnectorStatusId.ACTIVE)
+        
+        switch((dapsRegistrationSuccess ?? false, connectorStatus))
         {
-            await DeleteUpdateConnectorDetail(connectorId, iamUserId, cancellationToken, DapsClientId, connectorsRepository);
+            case (true, ConnectorStatusId.ACTIVE) when selfDescriptionDocumentId == null:
+                await DeleteUpdateConnectorDetail(connectorId, iamUserId, cancellationToken, dapsClientId, connectorsRepository);
+                break;
+            case (true, ConnectorStatusId.ACTIVE) when selfDescriptionDocumentId != null:
+                await DeleteConnector(connectorId, iamUserId, cancellationToken, dapsClientId, selfDescriptionDocumentId.Value, documentStatus!.Value, connectorsRepository);
+                break;
+            case (false, ConnectorStatusId.PENDING) when selfDescriptionDocumentId == null:
+                await DeleteConnectorWithoutDocuments(connectorId, connectorsRepository);
+                break;
+            case (false, ConnectorStatusId.PENDING) when selfDescriptionDocumentId != null:
+                await DeleteConnectorWithDocuments(connectorId, selfDescriptionDocumentId.Value, connectorsRepository);
+                break;
+            default:
+                throw new ConflictException($"Connector status does not match a deletion scenario. Deletion declined");
         }
-        else if ((SelfDescriptionDocumentid != null) && !DapsRegistrationSuccess!.Value && ConnectorStatus == ConnectorStatusId.PENDING)
-        {
-            await DeleteConnectorWithDocuments(connectorId, SelfDescriptionDocumentid, connectorsRepository);
-        }
-        else if ((SelfDescriptionDocumentid != null) && DapsRegistrationSuccess!.Value && ConnectorStatus == ConnectorStatusId.ACTIVE)
-        {
-            await DeleteConnector(connectorId, iamUserId, cancellationToken, DapsClientId, SelfDescriptionDocumentid, DocumentStatus, connectorsRepository);
-        }
-        else if ((SelfDescriptionDocumentid == null) && !DapsRegistrationSuccess!.Value && ConnectorStatus == ConnectorStatusId.PENDING)
-        {
-            await DeleteConnectorWithoutDocuments(connectorId, connectorsRepository);
-        }
-        else
-        {
-            throw new ConflictException($"Connector status does not match a deletion scenario. Deletion declined");
-        }
+
     }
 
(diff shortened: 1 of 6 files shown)


src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -305,28 +305,51 @@ public async Task DeleteConnectorAsync(Guid connectorId, string iamUserId, Cance
...
+        var (IsConnectorIdExist, DapsClientId, SelfDescriptionDocumentid, DocumentStatus, ConnectorStatus, DapsRegistrationSuccess) = result;
+        if (!IsConnectorIdExist)
         {
             throw new NotFoundException($"Connector {connectorId} does not exist");
         }
-
-        if (result.ConnectorStatus == ConnectorStatusId.INACTIVE)
+        if ((SelfDescriptionDocumentid == null) && DapsRegistrationSuccess!.Value && ConnectorStatus == ConnectorStatusId.ACTIVE)
         {
-            throw new ConflictException("INACTIVE Connector can not be deleted");
```
jjeroch, line 315: @qxz2mqe is this exception now fully gone? I would still expect the exception to be thrown, in case the connector send for deletion is already in Status "INACTIVE"

jjeroch (changes requested):

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
qxz2mqe: Julia this condition will not be met now because we have already added the condition that connector statusid is actve then only execute the existing implementation. i.e DAPS=true, SD=false and 
connector status = ACTIVE

jjeroch: So what happens if the endpoint is getting triggered by an user where the connector has the status "INACTIVE"?

qxz2mqe: In that case it will throw  throw new ConflictException($"Connector status does not match a deletion scenario. Deletion declined");

jjeroch: ok thats fine.
@qxz2mqe you can make this more transparent in future, if you add such information in your first feedback (the feedback which you added three days ago). You can just state more details such as:
`"the error scenario "throw new ConflictException("INACTIVE Connector can not be deleted")" is already covered by the newly added exception "throw new ConflictException($"Connector status does not match a deletion scenario. Deletion declined")" `
That would make it directly transparent

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -305,28 +305,51 @@ public async Task DeleteConnectorAsync(Guid connectorId, string iamUserId, Cance
...
+        else
+        {
+            throw new ConflictException($"Connector status does not match a deletion scenario. Deletion declined");
+        }
+    }
 
-        if (result.SelfDescriptionDocumentId != null)
+    private async Task DeleteConnector(Guid connectorId, string iamUserId, CancellationToken cancellationToken, string? DapsClientId, Guid? SelfDescriptionDocumentid, DocumentStatusId? DocumentStatus, IConnectorsRepository connectorsRepository)
+    {
+        if (SelfDescriptionDocumentid != null)
```
ntruchsess, line 337: define selfDescriptionDocumentId as being non-nullable in the methods signature. That allows to remove this non-null check. The only place where this method is being called does validate it's non null before the call.

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -305,28 +305,51 @@ public async Task DeleteConnectorAsync(Guid connectorId, string iamUserId, Cance
...
         {
             _portalRepositories.GetInstance<IDocumentRepository>().AttachAndModifyDocument(
-                result.SelfDescriptionDocumentId.Value,
-                a => { a.DocumentStatusId = result.DocumentStatusId!.Value; },
+                SelfDescriptionDocumentid.Value,
+                a => { a.DocumentStatusId = DocumentStatus!.Value; },
```
ntruchsess, line 341: technically it's incorrect to use the !-operator here as there is no validation documentStatus is actually non-null. Please define documentStatus as non-nullable in the method signature and implement the required handling of null value in the calling code. I suggest to throw UnexpectedConditionException in case documentStatus is null while selfDesctiptionDocumentId is non-null. (If this ever happens it means either there is a bug in the query or the database-scheme had changed such that the query is no longer valid and needs to be adjusted).

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -305,28 +305,51 @@ public async Task DeleteConnectorAsync(Guid connectorId, string iamUserId, Cance
...
                 a => { a.DocumentStatusId = DocumentStatusId.INACTIVE; });
         }
+        await DeleteUpdateConnectorDetail(connectorId, iamUserId, cancellationToken, DapsClientId, connectorsRepository);
+    }
+
+    private async Task DeleteUpdateConnectorDetail(Guid connectorId, string iamUserId, CancellationToken cancellationToken, string? DapsClientId, IConnectorsRepository connectorsRepository)
```
ntruchsess, line 347: DapsClientId -> lowerCase

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -305,28 +305,51 @@ public async Task DeleteConnectorAsync(Guid connectorId, string iamUserId, Cance
...
+            await DeleteConnectorWithoutDocuments(connectorId, connectorsRepository);
         }
+        else
+        {
+            throw new ConflictException($"Connector status does not match a deletion scenario. Deletion declined");
+        }
+    }
 
-        if (result.SelfDescriptionDocumentId != null)
+    private async Task DeleteConnector(Guid connectorId, string iamUserId, CancellationToken cancellationToken, string? DapsClientId, Guid? SelfDescriptionDocumentid, DocumentStatusId? DocumentStatus, IConnectorsRepository connectorsRepository)
```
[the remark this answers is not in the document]
qxz2mqe: resolved

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
qxz2mqe: resolved

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
qxz2mqe: resolved

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
qxz2mqe: resolved

ntruchsess (approved):

ntruchsess:
@Phil91 : I did rebase on latest dev. Can you please double check both my and @qxz2mqe 's changes

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -320,7 +342,20 @@ public async Task DeleteConnectorAsync(Guid connectorId, Guid userId, Cancellati
...
             con.DateLastChanged = DateTimeOffset.UtcNow;
         });
-        await _dapsService.DeleteDapsClient(result.DapsClientId, cancellationToken).ConfigureAwait(false);
+        await _dapsService.DeleteDapsClient(dapsClientId, cancellationToken).ConfigureAwait(false);
+        await _portalRepositories.SaveAsync();
+    }
+
+    private async Task DeleteConnectorWithDocuments(Guid connectorId, Guid selfDescriptionDocumentId, IConnectorsRepository connectorsRepository)
+    {
+        _portalRepositories.GetInstance<IDocumentRepository>().RemoveDocument(selfDescriptionDocumentId!);
```
Phil91, line 351: The `!` can be removed since the parameter isn't nullable

```suggestion
        _portalRepositories.GetInstance<IDocumentRepository>().RemoveDocument(selfDescriptionDocumentId);
```

Phil91 (changes requested):
one minor cosmetic finding

PR #113, opened 2023-06-28
chore: migrations since 1.5.0-rc1 have been merged into one

Phil91:
## Description

Merged all migrations since 1.5.0-rc1 to one migration

## Why

To have a better overview of the existing migrations on int

## Issue

N/A

## Checklist


evegufy (approved):

src/portalbackend/PortalBackend.Migrations/Migrations/PortalDbContextModelSnapshot.cs
```diff
@@ -5024,7 +5024,6 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasName("pk_verified_credential_external_type_use_case_detail_versions");
 
                     b.HasIndex("VerifiedCredentialExternalTypeId", "Version")
-                        .IsUnique()
```
ntruchsess, line 5027: This should not happen when aggregating migrations the final snapshot must not change.

ntruchsess (changes requested):

src/portalbackend/PortalBackend.Migrations/Migrations/PortalDbContextModelSnapshot.cs
Phil91: agreed, but seems like we had this one missing.

i double checked by creating an empty migration on the dev branch

ntruchsess (approved):

PR #120, opened 2023-07-03
feat(bpdm): add error handling for fetch legal entity 

Phil91:
## Description

Changed the url to fetch the legal entity, for the bpn check in the application process

## Why

The endpoint got enhanced by an errorMessage for the internal error handling. With the new endpoint an operator is able to see the specific error that occured and handle it.

## Issue

N/A - Jira Issue: CPLP-2143

## Checklist


src/externalsystems/Bpdm.Library/BpdmService.cs
```diff
@@ -122,27 +121,41 @@ public async Task<bool> PutInputLegalEntity(BpdmTransferData data, CancellationT
...
-                _options,
+                Options,
                 cancellationToken: cancellationToken).ConfigureAwait(false);
-            if (legalEntityResponse?.ExternalId == null)
+            if (paginationResponse == null || (paginationResponse.Content == null && paginationResponse.Errors == null))
+            {
+                throw new ServiceException("Access to external system bpdm did not return a valid legal entity response", true);
+            }
+
+            if (paginationResponse.Errors.Any())
```
ntruchsess, line 148: use 'IfAny()':
```
            paginationResponse.Errors.IfAny(errors =>
                throw new ServiceException($"The external system bpdm responded with errors {string.Join(";", errors.Select(x => $"ErrorCode: {x.ErrorCode}, ErrorMessage: {x.Message}"))}"));
```

src/externalsystems/Bpdm.Library/BpdmService.cs
```diff
@@ -122,27 +121,41 @@ public async Task<bool> PutInputLegalEntity(BpdmTransferData data, CancellationT
...
         try
         {
-            var legalEntityResponse = await JsonSerializer.DeserializeAsync<BpdmLegalEntityData>(
+            var paginationResponse = await JsonSerializer.DeserializeAsync<PageOutputResponseBpdmLegalEntityData>(
                 responseStream,
-                _options,
+                Options,
                 cancellationToken: cancellationToken).ConfigureAwait(false);
-            if (legalEntityResponse?.ExternalId == null)
+            if (paginationResponse == null || (paginationResponse.Content == null && paginationResponse.Errors == null))
```
ntruchsess, line 143: the check of Content and Errors seems to be redundant as both are defined as 'non-nullable' in BpdmLegalEntityOutputData.cs

src/externalsystems/Bpdm.Library/BpdmService.cs
```diff
@@ -122,27 +121,41 @@ public async Task<bool> PutInputLegalEntity(BpdmTransferData data, CancellationT
...
+                throw new ServiceException("Access to external system bpdm did not return a valid legal entity response", true);
+            }
+
+            if (paginationResponse.Errors.Any())
             {
-                throw new ServiceException("Access to external system bpdm did not return a valid legal entity response");
+                throw new ServiceException($"The external system bpdm responded with errors {string.Join(";", paginationResponse.Errors.Select(x => $"ErrorCode: {x.ErrorCode}, ErrorMessage: {x.Message}"))}");
             }
+
+            if (paginationResponse.Content!.Count(x => x.ExternalId == externalId) != 1)
```
ntruchsess, line 153: why not:
```
            try
            {
                return paginationResponse.Content!.Single(x => x.ExternalId == externalId);
            }
            catch(InvalidOperationException)
            {
                throw new ServiceException("Access to external system bpdm did not return a valid legal entity response", true);
            }
```

src/externalsystems/Bpdm.Library/BusinessLogic/IBpdmBusinessLogic.cs
```diff
@@ -18,6 +18,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
+using Org.Eclipse.TractusX.Portal.Backend.PortalBackend.PortalEntities.Enums;
```
ntruchsess, line 21: using Enums is not referenced and can be removed

src/externalsystems/Bpdm.Library/IBpdmService.cs
```diff
@@ -19,6 +19,8 @@
  ********************************************************************************/
 
 using Org.Eclipse.TractusX.Portal.Backend.Bpdm.Library.Models;
+using Org.Eclipse.TractusX.Portal.Backend.PortalBackend.PortalEntities.Enums;
```
ntruchsess, line 22: both usings (Enums + Library) are not referenced and shall be removed

ntruchsess (changes requested):

src/externalsystems/Bpdm.Library/BpdmService.cs
Phil91: the problem is that somehow the json serializer completly ignores the fact that the two values are not nullable and sets them to null. there is now jsonException thrown or anything. I therefor decided to mark the two properties as nullable and do the check.

ntruchsess (approved):

PR #121, opened 2023-07-04
feat(app)!: change app subscription process

Phil91:
## Description

an inactive subscription now won't be activated, instead a new subscription will be created

the following endpoints have been changed from appId to subscriptionId:

/api/apps/supscription/{subscriptiondId}/activate
/api/apps/{subscriptionId}/unsubscribe

## Why

Since the subscription is linked to a technical user and an app instance we don't want to reactivate an inactive subscription. Instead we want to create a new one.

## Issue

N/A - Jira Issue: CPLP-2911

## Checklist

### DIFF
--- a/src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
+++ b/src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
@@ -181,7 +181,6 @@ public class AppsBusinessLogic : IAppsBusinessLogic
         _offerSubscriptionService.AddOfferSubscriptionAsync(appId, offerAgreementConsentData, identity, OfferTypeId.APP, _settings.BasePortalAddress);
 
     /// <inheritdoc/>
-    [Obsolete("This Method is not used anymore")]
     public async Task ActivateOwnCompanyProvidedAppSubscriptionAsync(Guid subscriptionId, (Guid UserId, Guid CompanyId) identity)
     {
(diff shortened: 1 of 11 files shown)


src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
```diff
@@ -242,22 +237,31 @@ public async Task ActivateOwnCompanyProvidedAppSubscriptionAsync(Guid appId, Gui
     }
 
     /// <inheritdoc/>
-    public async Task UnsubscribeOwnCompanyAppSubscriptionAsync(Guid appId, Guid companyId)
+    public async Task UnsubscribeOwnCompanyAppSubscriptionAsync(Guid subscriptionId, Guid companyId)
```
ntruchsess, line 240: this method is lacking unit-tests

ntruchsess (changes requested):
I did rebase and change some exceptionTypes. Please pull before proceeding.

src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
Phil91: added unit tests

ntruchsess (approved):

ntruchsess:
We need to clarify whether to remove the 'obsolete'-annotations on the updated endpoints (and code being referenced)

PR #122, opened 2023-07-05
feat(logging): fix logging configuration

Phil91:
## Description

fixed the logging configuration where the logger had to be instantiated multiple times 
removed the health-check paths from request logging

## Why

the logs were spammed with the health check requests.

## Issue

N/A

## Checklist


src/framework/Framework.Web/WebApplicationBuildRunner.cs
```diff
@@ -37,7 +39,23 @@ public static class WebApplicationBuildRunner
         try
         {
             var builder = WebApplication.CreateBuilder(args);
-            builder.Host.AddLogging(builder.Configuration);
+
+            void ExtendLoggingForWeb(LoggerConfiguration configuration, IConfiguration config)
```
ntruchsess, line 43: as this private method is not being reused and there seems to be no apparent technical reason to implement as such I suggest passing this as lambda like so:
```
            builder.Host.AddLogging((LoggerConfiguration configuration, IConfiguration config) =>
            {
                configuration.Enrich.WithCorrelationIdHeader("X-Request-Id");
                var healthCheckPaths = config.GetSection("HealthChecks").Get<IEnumerable<HealthCheckSettings>>()?.Select(x => x.Path);
                if (healthCheckPaths != null)
                {
                    configuration
                        .Filter.ByExcluding(le =>
                        {
                            return le.Properties.TryGetValue("RequestPath", out var logProperty) &&
                                   logProperty.ToKeyValuePairs().Any(x => healthCheckPaths.Contains(x.Value));
                        });
                }
            });
```

ntruchsess (changes requested):

src/framework/Framework.Web/WebApplicationBuildRunner.cs
Phil91: agree. changed it to a lambda

PR #129, opened 2023-07-07
feat(connector)!: link managed connectors with offer subscription

Phil91:
## Description

change request /api/administration/connectors/managed-daps to take a subscriptionId instead of the providerBpn
new mapping table to link connectors with offer subscriptions
add database view to see all offer subscription related links

## Why

With this change the validation of the creation of a managed connector is enhanced to not only check if the customer is already an CX Member, additionally we check if the subscription is linked to an app which the caller provides.

## Issue

N/A - Jira Issue: CPLP-2831

## Checklist

❗  BREAKING CHANGE: The request of /api/administration/connectors/managed-daps has changed


src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -225,7 +248,8 @@ private async Task ValidateTechnicalUser(Guid? technicalUserId, Guid companyId)
         string businessPartnerNumber,
         Guid selfDescriptionDocumentId,
         IFormFile? file,
-        Guid? userId,
+        Guid companyUserId,
```
ntruchsess, line 251: the naming of the parameter seems to be misleading. From what I can see this can be a service-account as well

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -163,37 +163,60 @@ private async Task<Guid> CreateConnectorInternalAsync(ConnectorInputModel connec
...
+
+        if (!result.IsOfferProvider)
         {
-            throw new ControllerArgumentException($"Company {providerBpn} does not exist", nameof(providerBpn));
+            throw new ForbiddenException("Company is not the provider of the offer");
+        }
+
+        if (result.OfferSubscriptionAlreadyLinked)
+        {
+            throw new UnexpectedConditionException("OfferSubscription is already linked to a connector");
```
ntruchsess, line 191: ConflictException is more appropriate to indicate data in the database is not consistent with the current process-step. UnexpectedConditionException shall be used where code that is outside the local scope is expected to guarantee the fulfillment of a certain requirement (which might not (or no longer) be the case if that code changes or has a bug or the method is being used in a different context).

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -163,37 +163,60 @@ private async Task<Guid> CreateConnectorInternalAsync(ConnectorInputModel connec
...
+        }
+
+        if (result.OfferSubscriptionStatus != OfferSubscriptionStatusId.ACTIVE &&
+            result.OfferSubscriptionStatus != OfferSubscriptionStatusId.PENDING)
+        {
+            throw new UnexpectedConditionException($"The offer subscription must be either {OfferSubscriptionStatusId.ACTIVE} or {OfferSubscriptionStatusId.PENDING}");
```
ntruchsess, line 197: ConflictException

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -163,37 +163,60 @@ private async Task<Guid> CreateConnectorInternalAsync(ConnectorInputModel connec
...
         }
 
         if (result.SelfDescriptionDocumentId is null)
         {
             throw new UnexpectedConditionException($"provider company {result.CompanyId} has no self description document");
         }
+
+        if (string.IsNullOrWhiteSpace(result.ProviderBpn))
+        {
+            throw new UnexpectedConditionException($"The bpn of compay {result.CompanyId} must be set");
```
ntruchsess, line 207: ConflictException

ntruchsess (changes requested):

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
Phil91: changed

ntruchsess (dismissed):

ntruchsess (approved):

PR #135, opened 2023-07-12
fix(notification): get notification details

qxz2mqe:
## Description

Changed NotificationTopicId to nullable, to make sure the database query doesn't throw an exception when a link for notificationTypeId and NotificationTopicId is missing

## Why

Endpoint was throwing exception if there was no link data for notificationtypeid and notificationtopicid

## Issue

N/A - Jira Ticket: CPLP-2934

## Checklist


Phil91 (changes requested):
Please add a unit test for the repository method that is used for the endpoint to prove that the query does not fail if the notificationTopicId not assigned to an notificationTypeId.

The Unit test should be kind of the same as GetCountDetailsForUserAsync_ReturnsExpectedCount

qxz2mqe:
@Phil91  Added Unit Test

jjeroch:
@qxz2mqe  can you please check the "description"  and "why" of the pr - seems to be a draft content

qxz2mqe:
@jjeroch  Updated the description but I don't understand Draft Content??

Phil91 (approved):

PR #145, opened 2023-07-17
feat(sa): enhance get service accounts by Id

qxz2mqe:
## Description

added validation for company of user executing endpoint is either owner or provider

## Why

Earlier check was If user belongs to same company then allow to execute the end point now if it is of owner or provider company also then it should allow to execute the endpoint

## Issue

CPLP-2830

## Checklist


src/administration/Administration.Service/BusinessLogic/ServiceAccountBusinessLogic.cs
```diff
@@ -137,9 +137,16 @@ public async Task<int> DeleteOwnCompanyServiceAccountAsync(Guid serviceAccountId
     }
 
     public async Task<ServiceAccountConnectorOfferData> GetOwnCompanyServiceAccountDetailsAsync(Guid serviceAccountId, Guid companyId)
+    {
+        if (!await _portalRepositories.GetInstance<IServiceAccountRepository>().IsCompanyServiceAccountLinkedCompany(serviceAccountId, companyId).ConfigureAwait(false))
```
Phil91, line 141: If executing this with an technical user the first if statement will be successful, but the result returned from `GetOwnCompanyServiceAccountDetailedDataUntrackedAsync` will be null and a ConflictException will be thrown.

Please remove the `IsCompanyServiceAccountLinkedCompany` call and edit the `GetOwnCompanyServiceAccountDetailedDataUntrackedAsync` 

Please test the changes locally before rerequesting the review

Phil91 (changes requested):

src/portalbackend/PortalBackend.DBAccess/Repositories/ServiceAccountRepository.cs
```diff
@@ -115,7 +115,7 @@ public ServiceAccountRepository(PortalDbContext portalDbContext)
             .Where(serviceAccount =>
                 serviceAccount.Id == serviceAccountId &&
                 serviceAccount.Identity!.UserStatusId == UserStatusId.ACTIVE &&
-                serviceAccount.Identity.CompanyId == companyId)
+                serviceAccount.Identity.CompanyId == companyId && (serviceAccount.CompaniesLinkedServiceAccount!.Owners == companyId || serviceAccount.CompaniesLinkedServiceAccount!.Provider == companyId))
```
Phil91, line 118: The check of the identity.CompanyId can be removed since we do the check for a matching company with the companiesLinkedServiceAccount

```suggestion
                serviceAccount.CompaniesLinkedServiceAccount!.Owners == companyId || serviceAccount.CompaniesLinkedServiceAccount!.Provider == companyId)
```

Phil91 (changes requested):
Please add some UnitTests for `GetOwnCompanyServiceAccountDetailedDataUntrackedAsync` to check the result for the company check

src/portalbackend/PortalBackend.DBAccess/Repositories/ServiceAccountRepository.cs
qxz2mqe: fixed

Phil91 (changes requested):
The `ServiceAccountRepositoryTests` should include additional two Tests to check the result of the `GetOwnCompanyServiceAccountDetailedDataUntrackedAsync` one if the serviceAccount has different OwnerId than ProviderId the companyId equals the provider of the serviceAccount. And one to test if we have a valid serviceAccountId with a not matching companyId

qxz2mqe:
@Phil91  Added the Unit Test

Phil91 (approved):

Phil91 (approved):

PR #148, opened 2023-07-18
feat: add checks for serviceAccount and companyUsers to endpoints

Phil91:
## Description

add check whether an endpoint should only be callable for a service account user
add check whether an endpoint should only be callable for a company user

## Why

To enhance the endpoint security for endpoints that either are only callable by a company user or by a service account user

## Issue

N/A - Jira Ticket: CPLP-2863

## Checklist


src/administration/Administration.Service/Controllers/CompanyDataController.cs
```diff
@@ -278,6 +278,7 @@ public async Task<NoContentResult> CreateSsiCertificate([FromForm] SsiCertificat
     [HttpPut]
     [Authorize(Roles = "decision_ssicredential")]
     [Authorize(Policy = PolicyTypes.ValidIdentity)]
+    [Authorize(Policy = PolicyTypes.CompanyUser)]
```
ntruchsess, line 281: We should discuss whether  to ease maintenance - we want to define as little policies as possible (like PolicyTypes.CompanyUser would imply ValidIdentity). Now we have endpoints that combine those 2 and some that don't

ntruchsess (changes requested):

ntruchsess (approved):

PR #152, opened 2023-07-19
feat(apps & services): put and post enpoints

qxz2mqe:
## Description

Added DateLastChanged for most of endpoints of apps and services for put and post
## Why

Added DateLastChanged so that whenever we are updating or creating a record for offer related table there should be update for last date 
## Issue

CPLP-2886

## Checklist

- [ ] I have checked that new and existing tests pass locally with my changes
- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
+++ b/src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
@@ -147,6 +147,8 @@ public class AppsBusinessLogic : IAppsBusinessLogic
     public async Task AddFavouriteAppForUserAsync(Guid appId, Guid userId)
     {
         _portalRepositories.GetInstance<IOfferRepository>().CreateAppFavourite(appId, userId);
+        _portalRepositories.GetInstance<IOfferRepository>().AttachAndModifyOffer(appId, offer =>
+            offer.DateLastChanged = DateTimeOffset.UtcNow);
         await _portalRepositories.SaveAsync().ConfigureAwait(false);
     }
 
--- a/src/marketplace/Services.Service/BusinessLogic/ServiceReleaseBusinessLogic.cs
+++ b/src/marketplace/Services.Service/BusinessLogic/ServiceReleaseBusinessLogic.cs
@@ -198,10 +198,12 @@ public class ServiceReleaseBusinessLogic : IServiceReleaseBusinessLogic
                 offer.SalesManagerId = data.SalesManager;
                 offer.ContactEmail = data.ContactEmail;
                 offer.MarketingUrl = data.ProviderUri;
+                offer.DateLastChanged = DateTimeOffset.UtcNow;
             },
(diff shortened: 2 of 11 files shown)


src/marketplace/Services.Service/BusinessLogic/ServiceReleaseBusinessLogic.cs
```diff
@@ -198,10 +198,12 @@ public async Task UpdateServiceAsync(Guid serviceId, ServiceUpdateRequestData da
                 offer.SalesManagerId = data.SalesManager;
                 offer.ContactEmail = data.ContactEmail;
                 offer.MarketingUrl = data.ProviderUri;
+                offer.DateLastChanged = DateTimeOffset.UtcNow;
```
Phil91, line 201: please don't set the DateLastChanged in the initialize to DateTimeOffset.UtcNow, either set it to the correct database value or remove the call since the value will be set in the modify action and the set value isn't null or default

Phil91 (changes requested):
the DateLastChanged is currently not set within the 

/api/apps/AppReleaseProcess/createapp endpoint, please add that to the logic

Phil91:
@jjeroch  most of the endpoints are not directly changing the offer. in most cases we set or remove an entry in a mapping table.

exceptions for that are:

/api/apps/AppReleaseProcess/createapp
/api/apps/AppReleaseProcess/{appId}

do we really want to change the logic this way?

src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
```diff
@@ -147,6 +147,8 @@ public async Task RemoveFavouriteAppForUserAsync(Guid appId, Guid userId)
     public async Task AddFavouriteAppForUserAsync(Guid appId, Guid userId)
     {
         _portalRepositories.GetInstance<IOfferRepository>().CreateAppFavourite(appId, userId);
+        _portalRepositories.GetInstance<IOfferRepository>().AttachAndModifyOffer(appId, offer =>
```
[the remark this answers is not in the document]
qxz2mqe: fixed

src/marketplace/Services.Service/BusinessLogic/ServiceReleaseBusinessLogic.cs
qxz2mqe: fixed

Phil91:
> @jjeroch most of the endpoints are not directly changing the offer. in most cases we set or remove an entry in a mapping table.
> 
> exceptions for that are:
> 
> /api/apps/AppReleaseProcess/createapp /api/apps/AppReleaseProcess/{appId}
> 
> do we really want to change the logic this way?

clarified, the logic should be as soon as an offer or offer related entity changes, the DateLastChanged of the offer should be updated

Phil91 (approved):

PR #161, opened 2023-07-21
feat(openUrl): Add Endpoints for public apis foreach service

Phil91:
## Description

Add GET endpoint to receive public apis foreach service

## Why

To be able to display the public available urls in the frontend a new endpoint was needed

## Issue

N/A - Jira Issue: CPLP-2958

## Checklist


src/portalbackend/PortalBackend.DBAccess/Repositories/CompanyRepository.cs
```diff
@@ -290,6 +290,13 @@ public void AttachAndModifyProviderCompanyDetails(Guid providerCompanyDetailId,
             ))
             .SingleOrDefaultAsync();
 
+    /// <inheritdoc />
+    public Task<CompanyRoleId[]> GetOwnCompanyRolesAsync(Guid companyId) =>
+        _context.CompanyAssignedRoles
+            .Where(x => x.CompanyId == companyId)
+            .Select(x => x.CompanyRoleId)
+            .ToArrayAsync();
```
ntruchsess, line 298: the query should return IAsyncEnumerable, any materialization via ToArrayAsync or ToListAsync should be executed in the businessLogic

src/framework/Framework.PublicInfos/PublicInformationBusinessLogic.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// </summary>
+    /// <param name="actionDescriptorCollectionProvider">The actionDescriptorCollectionProvider</param>
+    /// <param name="portalRepositories"></param>
+    public PublicInformationBusinessLogic(IActionDescriptorCollectionProvider actionDescriptorCollectionProvider, IPortalRepositories portalRepositories)
+    {
+        _actionDescriptorCollectionProvider = actionDescriptorCollectionProvider;
+        _portalRepositories = portalRepositories;
+    }
+
+    public async Task<IEnumerable<UrlInformation>> GetPublicUrls(Guid companyId)
```
ntruchsess, line 47: 1) I suggest to use linq instead of list + foreach/add:
```
    public async Task<IEnumerable<UrlInformation>> GetPublicUrls(Guid companyId)
    {
        var companyRoleIds = await _portalRepositories.GetInstance<ICompanyRepository>().GetOwnCompanyRolesAsync(companyId).ConfigureAwait(false);
        return _actionDescriptorCollectionProvider.ActionDescriptors.Items
            .Where(item => item.ActionConstraints != null)
            .OfType<ControllerActionDescriptor>()
            .Where(item => item.MethodInfo.GetCustomAttribute<PublicUrlAttribute>()?.CompanyRoleIds.Intersect(companyRoleIds).Any() ?? false)
            .Select(item =>
                new UrlInformation(
                    string.Join(", ", item.ActionConstraints!.OfType<HttpMethodActionConstraint>().SelectMany(x => x.HttpMethods).Distinct()),
                    (item.AttributeRouteInfo?.Template ?? throw new ConflictException($"There must be an url for {item.DisplayName}")).ToLower()));
    }
```
2) this method is missing an explicit unit-test that works independent from real controller classes (using a faked IActionDescriptorCollection + faked repository-call)

ntruchsess (changes requested):

src/portalbackend/PortalBackend.DBAccess/Repositories/CompanyRepository.cs
Phil91: adjusted

src/framework/Framework.PublicInfos/PublicInformationBusinessLogic.cs
Phil91: changed it to a linq statement instead and added unit tests for it.

tests/maintenance/Maintenance.App.Tests/Setup/TestDbFixture.cs
```diff
@@ -96,7 +95,7 @@ await _container.StartAsync()
         );
         var context = new PortalDbContext(optionsBuilder.Options, new FakeIdentityService());
         await context.Database.MigrateAsync();
-        BaseSeed.SeedBasedata().Invoke(context);
+        // BaseSeed.SeedData().Invoke(context);
```
ntruchsess, line 98: that commented line of code shall be removed. Looks like the previously called method now does not have any references and can be removed as well. (please also check the other static 'SeedData()'-methods as the appear to also have 0 references, but maybe my code-scanner is incorrect here?)

ntruchsess (changes requested):

tests/maintenance/Maintenance.App.Tests/Setup/TestDbFixture.cs
Phil91: removed the commented out code.

The SeedData is used in `ServiceCollectionExtensions`  in the Method `EnsureDbCreatedWithSeeding<TSeedingData>`

ntruchsess (approved):

PR #170, opened 2023-07-26
fix(notification): link creatorId to identity

Phil91:
## Description

The database foreign key of creatorId is not linked to a companyUser anymore, instead it's now linked to a identity

## Why

Previously the creator of a notification was linked to a companyUser which results in errors when executing the autoSetup with a technicalUser, thats fixed with this pr

## Issue

N/A - Jira Issue: CPLP-3042

## Checklist


src/portalbackend/PortalBackend.Migrations/Migrations/PortalDbContextModelSnapshot.cs
```diff
@@ -1,4 +1,4 @@
-/********************************************************************************
+﻿/********************************************************************************
```
ntruchsess: please save as UTF-8 (without BOM)

ntruchsess (approved):

src/portalbackend/PortalBackend.Migrations/Migrations/PortalDbContextModelSnapshot.cs
ntruchsess: fixed

PR #250, opened 2023-09-04
fix(service) : added leadiamge id and datelastchanged fields

VPrasannaK94:
## Description

added leadiamge id and datelastchanged fields

## Why

Additional fields added to existing endpoint

## Issue

[CPLP-3169](https://jira.catena-x.net/browse/CPLP-3169)

## Checklist

### DIFF
--- a/src/marketplace/Services.Service/BusinessLogic/ServiceBusinessLogic.cs
+++ b/src/marketplace/Services.Service/BusinessLogic/ServiceBusinessLogic.cs
@@ -166,12 +166,7 @@ public class ServiceBusinessLogic : IServiceBusinessLogic
                     companyProvidedServiceStatusData.Data.Select(item =>
                         item with
                         {
-                            Id = item.Id,
-                            Name = item.Name,
-                            LeadPictureId = item.LeadPictureId == Guid.Empty ? null : item.LeadPictureId,
-                            Provider = item.Provider,
-                            Status = item.Status,
-                            LastChanged = item.LastChanged
+                            LeadPictureId = item.LeadPictureId == Guid.Empty ? null : item.LeadPictureId
                         }));
         }
         return await Pagination.CreateResponseAsync(page, size, 15, GetCompanyProvidedServiceStatusData).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


src/marketplace/Services.Service/BusinessLogic/ServiceBusinessLogic.cs
```diff
@@ -152,10 +152,28 @@ public async Task<Pagination.Response<OfferCompanySubscriptionStatusResponse>> G
...
+                    companyProvidedServiceStatusData.Count,
+                    companyProvidedServiceStatusData.Data.Select(item =>
+                        new AllOfferStatusData(
+                            item.Id,
+                            item.Name,
+                            item.LeadPictureId == Guid.Empty ? null : item.LeadPictureId,
+                            item.Provider,
+                            item.Status,
+                            item.LastChanged)));
+        }
```
Phil91, line 174: can be changed to:

```suggestion
                    companyProvidedServiceStatusData.Data.Select(item =>
                        item with { LeadPictureId = item.LeadPictureId == Guid.Empty ? null : item.LeadPictureId }));
        }
```

Phil91 (changes requested):

src/marketplace/Services.Service/BusinessLogic/ServiceBusinessLogic.cs
```diff
@@ -152,10 +152,30 @@ public async Task<Pagination.Response<OfferCompanySubscriptionStatusResponse>> G
...
+        {
+            var companyProvidedServiceStatusData = await _portalRepositories.GetInstance<IOfferRepository>()
+                .GetCompanyProvidedServiceStatusDataAsync(GetOfferStatusIds(statusId), OfferTypeId.SERVICE, companyId, sorting ?? OfferSorting.DateDesc, offerName)(skip, take).ConfigureAwait(false);
+
+            return companyProvidedServiceStatusData == null
+                ? null
+                : new Pagination.Source<AllOfferStatusData>(
+                    companyProvidedServiceStatusData.Count,
+                    companyProvidedServiceStatusData.Data.Select(item =>
+                        item with
```
Phil91, line 167: you only need to set the properties which aren't assigned to the same property of item. e.g. you don't need to set the item.Id because that will be automatically taken. In this case only the LeadPictureId must be set, because thats the only property the might be different to the item. one

Phil91 (changes requested):

Phil91 (approved):

PR #274, opened 2023-09-25
feat(kc): add configurable auth path for keycloak requests

Phil91:
## Description

Make the auth path of keycloak configureable

## Why

With the keycloak upgrade the /auth path in the urls of keycloak was removed, to be able to still use older versions its now configurable

## Issue

N/A - Jira Issue: CPLP-3214

## Depending CD PR
[#95](https://github.com/eclipse-tractusx/portal-cd/pull/95)

## Checklist

### DIFF
--- a/src/keycloak/Keycloak.Factory/KeycloakSettingData.cs
+++ b/src/keycloak/Keycloak.Factory/KeycloakSettingData.cs
@@ -27,19 +27,20 @@ namespace Org.Eclipse.TractusX.Portal.Backend.Keycloak.Factory;
 
 public class KeycloakSettings
 {
-    public KeycloakSettings()
-    {
-        ConnectionString = null!;
-    }
+    public string ConnectionString { get; set; } = null!;
 
-    public string ConnectionString { get; set; }
     public string? User { get; set; }
-    public string? Password { get; set; }
-    public string? ClientId { get; set; }
-    public string? ClientSecret { get; set; }
-    public string? AuthRealm { get; set; }
 
-    public void Validate(string key)
+    public string? Password { get; set; }
+
+    public string? ClientId { get; set; }
+
+    public string? ClientSecret { get; set; }
+
+    public string? AuthRealm { get; set; }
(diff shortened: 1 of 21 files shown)


src/keycloak/Keycloak.Factory/KeycloakSettingData.cs
```diff
@@ -27,19 +27,20 @@ namespace Org.Eclipse.TractusX.Portal.Backend.Keycloak.Factory;
...
     public string? User { get; set; }
+
     public string? Password { get; set; }
+
     public string? ClientId { get; set; }
+
     public string? ClientSecret { get; set; }
+
     public string? AuthRealm { get; set; }
+    public bool UseAuthTrail { get; set; }
```
evegufy, line 41: Looks good! could you please add the setting in every relevant appsettings.json file with default value 'false'.

evegufy (changes requested):

src/keycloak/Keycloak.Factory/KeycloakSettingData.cs
Phil91: sure, done.

evegufy: Thanks! I re-thought our conversation from earlier: if you create a PR for the helm chart adding this setting with value 'true', it shouldn't be breaking change for dev, could you please also create that PR?
Also there's a code smell, could you handle that?

Phil91: will take care about the pr.
regarding the code smell nothing really we could do against it, we need the parameters in the method. but i'll add a comment in sonar

Phil91: CD [PR](https://github.com/eclipse-tractusx/portal-cd/pull/95) has been created. I've added a comment for the code smell in sonar

evegufy (approved):

PR #315, opened 2023-10-26
feat(errorhandling)! enhance exceptions with details for errortype, errorcode and parameters

ntruchsess:
## Description

Exceptions in Framework.ErrorHandling have been enhanced. They now may contain an error-type, an error-code, a message-template and multiple parameters. This allows structured logging of error-details and a more precise display of errors in the frontend (and optionally internationalization by including multi-language error-message-resources for specific errorcodes into the frontend). As those details are optional this change is fully backwards-compatible and api-endpoints may be adjusted on an 'as needed' basis.

## Why

error-messages being returned from api-endpoints are just application-identifier (often a fully qualified class-name) plus string-message. This cannot easily be parsed by the frontend to map to more detailed (and optionally multilingual) help-texts. Parameters of methods throwing exceptions are embedded within the message and therefore not available explicitly.

## Issue

N/A (CPLP-3090)

## Checklist

- [X] I have followed the [contributing guidelines](https://github.com/eclipse-tractusx/portal-assets/blob/main/developer/Technical%20Documentation/Dev%20Process/How%20to%20contribute.md#commit-and-pr-guidelines)
- [X] I have added tests that prove my changes work
- [X] I have checked that new and existing tests pass locally with my changes

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
@@ -20,6 +20,7 @@
 
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Options;
+using Org.Eclipse.TractusX.Portal.Backend.Administration.ErrorHandling;
 using Org.Eclipse.TractusX.Portal.Backend.Administration.Service.Models;
 using Org.Eclipse.TractusX.Portal.Backend.Clearinghouse.Library.BusinessLogic;
 using Org.Eclipse.TractusX.Portal.Backend.Clearinghouse.Library.Models;
@@ -82,7 +83,7 @@ public sealed class RegistrationBusinessLogic : IRegistrationBusinessLogic
         var companyWithAddress = await _portalRepositories.GetInstance<IApplicationRepository>().GetCompanyUserRoleWithAddressUntrackedAsync(applicationId).ConfigureAwait(false);
         if (companyWithAddress == null)
         {
-            throw new NotFoundException($"applicationId {applicationId} not found");
+            throw NotFoundException.Create(AdministrationRegistrationErrors.APPLICATION_NOT_FOUND, new ErrorParameter[] { new("applicationId", applicationId.ToString()) });
         }
         return new CompanyWithAddressData(
             companyWithAddress.CompanyId,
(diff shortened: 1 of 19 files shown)


src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -82,7 +83,7 @@ private async Task<CompanyWithAddressData> GetCompanyWithAddressAsyncInternal(Gu
         var companyWithAddress = await _portalRepositories.GetInstance<IApplicationRepository>().GetCompanyUserRoleWithAddressUntrackedAsync(applicationId).ConfigureAwait(false);
         if (companyWithAddress == null)
         {
-            throw new NotFoundException($"applicationId {applicationId} not found");
+            throw NotFoundException.Create(AdministrationRegistrationErrors.APPLICATION_NOT_FOUND, new ErrorParameter[] { new("applicationId", applicationId.ToString()) });
```
Phil91, line 86: I'd suggest changing the ToString to nameof. Should result in the same

```suggestion
            throw NotFoundException.Create(AdministrationRegistrationErrors.APPLICATION_NOT_FOUND, new ErrorParameter[] { new("applicationId", nameof(applicationId)) });
```

src/framework/Framework.ErrorHandling.Web/GeneralHttpErrorHandler.cs
```diff
@@ -45,10 +47,11 @@ public class GeneralHttpErrorHandler
         { HttpStatusCode.InternalServerError, new MetaData("https://datatracker.ietf.org/doc/html/rfc7231#section-6.6.1", "The server encountered an unexpected condition.") }
     }.ToImmutableDictionary();
 
-    public GeneralHttpErrorHandler(RequestDelegate next, ILogger<GeneralHttpErrorHandler> logger)
+    public GeneralHttpErrorHandler(RequestDelegate next, ILogger<GeneralHttpErrorHandler> logger, IErrorMessageService? errorMessageService = null)
```
Phil91, line 50: I guess the errorMessageService is currently nullable so we don't have to adjust all the code at once and will be changed in the future to a not nullable?

If that's the case let's maybe add an todo here

Phil91 (changes requested):
One minor question open.
Besides of that looking really good.

Could you please comment the sonar findings - in my opinion that are false positives

src/framework/Framework.ErrorHandling.Web/GeneralHttpErrorHandler.cs
ntruchsess: yes, that is the reason. I did add a TODO as reminder

Phil91 (approved):

PR #321, opened 2023-11-01
fix(e2e): adjust end to end tests

Phil91:
## Description

Fix end to end tests

## Why

Currently the end to end tests are running on an error on dev as well as int

## Issue

N/A - Jira Ref: CPLP-3328

## Checklist

### DIFF
--- a/src/portalbackend/PortalBackend.DBAccess/Repositories/UserRolesRepository.cs
+++ b/src/portalbackend/PortalBackend.DBAccess/Repositories/UserRolesRepository.cs
@@ -116,9 +116,9 @@ public class UserRolesRepository : IUserRolesRepository
             .AsNoTracking()
             .Where(role =>
                 role.OfferId == offerId &&
-                role.UserRoleCollections.Any(collection => collection.CompanyRoleAssignedRoleCollection!.CompanyRole!.CompanyAssignedRoles.Any(assigned => assigned.Company!.Identities.Any(user => user.Id == companyUserId))) &&
-                (userRoles.Contains(role.UserRoleText) ||
-                role.IdentityAssignedRoles.Select(u => u.Identity!).Any(user => user.Id == companyUserId)))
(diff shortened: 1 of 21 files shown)


src/portalbackend/PortalBackend.DBAccess/Repositories/UserRolesRepository.cs
```diff
@@ -117,9 +117,9 @@ public async IAsyncEnumerable<Guid> GetUserRoleIdsUntrackedAsync(IEnumerable<Use
             .AsNoTracking()
             .Where(role =>
                 role.OfferId == offerId &&
-                role.UserRoleCollections.Any(collection => collection.CompanyRoleAssignedRoleCollection!.CompanyRole!.CompanyAssignedRoles.Any(assigned => assigned.Company!.Identities.Any(user => user.Id == companyUserId))) &&
-                (userRoles.Contains(role.UserRoleText) ||
-                role.IdentityAssignedRoles.Select(u => u.Identity!).Any(user => user.Id == companyUserId)))
+                (role.UserRoleCollections.Any(collection => collection.CompanyRoleAssignedRoleCollection!.CompanyRole!.CompanyAssignedRoles.Any(assigned => assigned.Company!.Identities.Any(user => user.Id == companyUserId))) ||
+                 role.IdentityAssignedRoles.Select(u => u.Identity!).Any(user => user.Id == companyUserId)) &&
```
ntruchsess, line 121: from my point of view this is incorrect. The query shall return all roles that belong to that offer and are eligible to be assigned according to the company's company-role that are either 'to-be-assigned' (parameter 'userRoles') **or** already assigned to the given user (plus the respective role's assignment-status)
The changed condition will return all roles that belong to the offer and are to be assigned while being either eligible to be assigned according to the company's company-role or already assigned to the given user. That may include roles that should not be assignable because they are not eligible to be assigned according to the company's company-roles. As a result roles that are already assigned but should no be assignable may stay assigned when updating the roles assignment.
(I also cannot see how this changes relates to the other content of this PR).

ntruchsess (changes requested):

ntruchsess (approved):

PR #334, opened 2023-11-09
fix(userrole): fix user role assignment

Phil91:
## Description

Adjusting the handling of user role assignments with multiple client ids.

## Why

Currently assigning roles to an offer with multiple client ids fails

## Issue

Refs: N/A - Jira Issue: TEST-1537

## Checklist

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/UserRolesBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/UserRolesBusinessLogic.cs
@@ -22,6 +22,7 @@ using Microsoft.Extensions.Options;
 using Org.Eclipse.TractusX.Portal.Backend.Administration.Service.Models;
 using Org.Eclipse.TractusX.Portal.Backend.Framework.Async;
 using Org.Eclipse.TractusX.Portal.Backend.Framework.ErrorHandling;
+using Org.Eclipse.TractusX.Portal.Backend.Framework.Linq;
 using Org.Eclipse.TractusX.Portal.Backend.Framework.Models;
 using Org.Eclipse.TractusX.Portal.Backend.PortalBackend.DBAccess;
 using Org.Eclipse.TractusX.Portal.Backend.PortalBackend.DBAccess.Models;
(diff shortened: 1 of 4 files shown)


src/administration/Administration.Service/BusinessLogic/UserRolesBusinessLogic.cs
```diff
@@ -210,9 +210,9 @@ private async Task<IEnumerable<UserRoleModificationData>> AddRoles(Guid companyU
     {
         var userRoleRepository = _portalRepositories.GetInstance<IUserRolesRepository>();
         var clientRoleNames = iamClientIds.ToDictionary(clientId => clientId, _ => rolesToAdd.Select(x => x.CompanyUserRoleText));
-        var assignedRoles = await _provisioningManager.AssignClientRolesToCentralUserAsync(iamUserId, clientRoleNames).SingleAsync().ConfigureAwait(false);
+        var assignedRoles = await _provisioningManager.AssignClientRolesToCentralUserAsync(iamUserId, clientRoleNames).ToListAsync().ConfigureAwait(false);
 
-        var rolesAdded = rolesToAdd.IntersectBy(assignedRoles.Roles, role => role.CompanyUserRoleText).ToList();
+        var rolesAdded = rolesToAdd.IntersectBy(assignedRoles.SelectMany(x => x.Roles).Distinct(), role => role.CompanyUserRoleText).ToList();
```
ntruchsess, line 215: rolesAdded will only contain the roles that have been assigned to **at least one** of the clients. In case a Role succeeds for one client but fails for a different one this will remain undetected.

To determine the roles that have been successfully assigned to **all** the clients I suggest:
```
var assignedRoles = _provisioningManager.AssignClientRolesToCentralUserAsync(iamUserId, clientRoleNames);

var rolesAdded = (await assignedRoles // determine the roles that have been assigned to all of the clients.
    .AggregateAsync(
        rolesToAdd,
        (IEnumerable<UserRoleModificationData> toAdd, (string ClientId, IEnumerable<string> Roles) assigned) =>
            toAdd.IntersectBy(assigned.Roles, role => role.CompanyUserRoleText)).ConfigureAwait(false))
    .ToList();
```

the expected behavious (fail early vs. ignore partial failure) should be clarified with @jjeroch

ntruchsess (changes requested):

src/administration/Administration.Service/BusinessLogic/UserRolesBusinessLogic.cs
Phil91: as discussed with @jjeroch we're using the fail early method.
code is adjusted accordingly

ntruchsess (approved):

PR #494, opened 2024-02-08
feat(errorhandling): add common controller returning all errormessages

ntruchsess:
## Description

a new endpoint which returns all error-types, error-codes and error-messages that a service makes use of has been added.
The implementation of this endpoint is common for all services. It retrieves the data dynamically from the services IErrorMessageService-implementation.

corresponding CD-Repository PR: https://github.com/eclipse-tractusx/portal-cd/pull/176

## Why

Currently no reliable documentation of error-types, error-codes and messages the services really make use of exists.

## Issue

N/A

## Checklist

- [X] I have followed the [contributing guidelines](https://github.com/eclipse-tractusx/portal-assets/blob/main/docs/developer/Technical%20Documentation/Dev%20Process/How%20to%20contribute.md#commit-and-pr-guidelines)


src/notifications/Notifications.Service/Properties/launchSettings.json
```diff
@@ -24,6 +24,7 @@
       "launchUrl": "api/notification/swagger",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
+        "MVC_ROUTING_BASEPATH": "api/Notification",
```
Phil91, line 27: could we please change that to lowercase, doesn't make a difference since it should be treated case insensitive but to have it similar to the other launch settings

tests/shared/Tests.Shared/IntegrationTests/IntegrationTestFactory.cs
```diff
@@ -55,6 +55,8 @@ protected override void ConfigureWebHost(IWebHostBuilder builder)
         var projectDir = Directory.GetCurrentDirectory();
         var configPath = Path.Combine(projectDir, "appsettings.IntegrationTests.json");
 
+        Environment.SetEnvironmentVariable("MVC_ROUTING_BASEPATH", "/api/test");
```
Phil91, line 58: Really love that we can set the urls to a specific value now!

Phil91 (changes requested):

Phil91 (commented):
One thing I just noticed, since we said we want to update all packages you might revert the changes of the existing Directory.Build.props and execute the update_framework_versions.sh script

src/notifications/Notifications.Service/Properties/launchSettings.json
ntruchsess: changed

ntruchsess:
> One thing I just noticed, since we said we want to update all packages you might revert the changes of the existing Directory.Build.props and execute the update_framework_versions.sh script

done

Issue #563, opened 2024-03-07
API GET Technical User(s) - filter on "isOwner" does not work as expected


jjeroch:
Detected in Version: **dev (not release candidate)**

In the current dev version, the endpoint
`/api/administration/serviceaccount/owncompany/serviceaccounts?size=10&page=0&isOwner=true`
(tested with CX Operator Login)
response with Managed Technical users; instead of only showing the once where the serviceAccountType is `"MANAGED"`

To be checked!

jjeroch:
@VPrasannaK94  & @ntruchsess 
I analyzed the situation based on the proposed pr fix once again.
The issue is not the filter itself; the issue appears on a much earlier state already. Its inside the technicalUserAccount creation.
When looking into the view portal.company_linked_service_accounts saw a number of accounts where following situation exists:
Owner ID and Provider ID is the same - still the technicalUserAccount is under the type "Managed".
I further validated the scenario and realized that this happens whenever a company subscribes to their own service/app offer.
To fix this we need to do the following:
- update technical user account creation backend logic of the following endpoints:
   - POST /api/apps/start-autoSetup
   - POST /api/services/start-autoSetup
in both the endpoints a technical user gets created if a technical user profile exists. This technical user account is always created as "MANAGED".  We need to enhance the logic to only create the technical user as "MANAGED" if the offer provider and subscription customer are not from the same company. Note: in 98% this will be the case; but I also like to have a clean solution for this other 2 percent.
If offer provider and subscription customer company ID is identical, the technicalUserAccount type should be "OWN".

As soon as this is done we need additionally a data fix script which runs with the migration/as part of the deployment.
Set all technicalUserAccount types to "OWN" where offer provider and subscription customer have the same companyID set.

jjeroch:
After discussion with @VPrasannaK94  & @ntruchsess suggestion is to implement the logic via the field "isOwner" of the GET endpoint - by enhancing the "isOwner" attribute logic with the following logic:
- if the portal.company_linked_service_accounts field provider is empty the "isOwner" flag is set to "true
- if the portal.company_linked_service_accounts field provider is filled, following needs to happen: 
  - if "provider" == "owner" set the "isOwner" flag to "true
  - otherwise "isOwner" flag to "false"

VPrasannaK94:
After discussion with @ntruchsess suggestion is to add one more field "isProvider" of the Get endpoint

[...]

PR #597, opened 2024-04-03
fix: fixed isOwner filter

VPrasannaK94:
## Description

fixed isOwner filter

## Why

when hitting GET: api/administration/serviceaccount/owncompany/serviceaccounts with isOwner flag as true it results
companyServiceAccountTypeIds as Managed Type records instead of showing only Own Type records, now it is fixed by adding additional check condition

## Issue

Ref : #563

## Checklist


ntruchsess (approved):

ntruchsess (changes requested):
As the ticket is not very clear about the expected semantics I did discuss with @jjeroch . We decided to put this PR on hold and check the semantics of view company_linked_service_account first.

jjeroch:
@ntruchsess  thanks.
I have validated the details and added the actual needed change in the ticket https://github.com/eclipse-tractusx/portal-backend/issues/563

ntruchsess (approved):

Issue #593, opened 2024-04-02
Technical User Creation in DIM Wallet via Portal Integration


jjeroch:
`Ticket under creation`

**Description**:
The current integration between our portal and the DIM Wallet requires enhancement to support the creation of technical users for holders. The existing service account creation service within the portal should be extended to include additional logic that triggers the creation of a technical user in the DIM Wallet when the "Wallet Identity Management" permission is granted.

**Acceptance Criteria**:
1. When a technical user is created in Keycloak with the "Wallet Identity Management" permission, the system should also create a corresponding technical user in the DIM Wallet.
2. The creation of the technical user in the DIM Wallet should be handled asynchronously to ensure the portal's performance is not adversely affected. (only if needed - synchron implementation would be preferred)
3. The system should handle any errors during the DIM Wallet technical user creation process gracefully and log them appropriately for troubleshooting.
4. The user interface on the portal should reflect the status of the DIM Wallet technical user creation process (e.g., pending, successful, failed). => @Phil91 lets check this requirement once
5. Make sure that the DIM Tech User has only limited permissions assigned to enable the communication of the Connector
6. Update any relevant documentation to reflect the new functionality and guide users on how to create a technical user with the "Wallet Identity Management" permission.
7. Implement necessary unit tests and integration tests to cover the new functionality.
8. Ensure that the existing functionality is not affected by these changes.

**Technical Notes**:
- The service account creation logic is currently managed by the `Administration Service` ServiceAccountController.
- The DIM Wallet API documentation provides the necessary endpoints for creating technical users and managing their permissions.
- The integration should use secure and authenticated communication channels to interact with the DIM Wallet API.

jjeroch:
Initial implementation done, but refactoring planned.
Suggestion to do this as part of this ticket since the solution is not working like we expected

PR #706, opened 2024-05-02
feat(dimTechnicalUser): adjust dim technical users

Phil91:
## Description

* add new provider information table
* add internal / external flag to service accounts

## Why

To support external company service accounts with the same client_client_id

## Issue

Refs: #593

## Corresponding CD PR

https://github.com/eclipse-tractusx/portal/pull/305

## Checklist

- [x] I have created and linked IP issues or requested their creation by a committer


src/portalbackend/PortalBackend.Migrations/Migrations/20240514145358_593-AdjustDimTechnicalUsers.cs
```diff
@@ -40,24 +42,77 @@ protected override void Up(MigrationBuilder migrationBuilder)
                 table: "company_service_accounts",
                 type: "integer",
                 nullable: false,
-                defaultValue: 1);
+                defaultValue: 0);
```
Phil91, line 45: I think we should change that to 1, think the migration will fail otherwise

ntruchsess: As company_service_account_kind_id is not meant to be optional I did add sql to the migration that updates the existing company_service_accounts setting the column to 1.

ntruchsess (approved):

PR #737, opened 2024-05-15
chore: improve dependabot and rework file header

evegufy:
## Description

dependabot group version updates
chore: rework year in file header

## Why

eclipse-tractusx/portal#300
eclipse-tractusx/portal#234

## Checklist

- [x] I have performed a self-review of my own changes


ntruchsess (approved):

ntruchsess (approved):

evegufy:
> @Phil91 I'm not sure how to proceed with the failing check for the nuget packages? https://github.com/eclipse-tractusx/portal-backend/actions/runs/9095882443/job/25000084520?pr=737 Can we ignore it in this case, or will that lead to issues after merge?

I decided to raise the patch version, just in case

Phil91:
> @Phil91 I'm not sure how to proceed with the failing check for the nuget packages? https://github.com/eclipse-tractusx/portal-backend/actions/runs/9095882443/job/25000084520?pr=737 Can we ignore it in this case, or will that lead to issues after merge?

theoretically this should not be a problem since we only edited the file header. but to be on the safe side we should upgrade the version of the packages

Issue #787, opened 2024-06-05
Provisioning after invitation fails | "Client saidp1 already exists"


evegufy:
### Current Behavior

It was noticed that sometimes (up to now only observed in [local dev setup](https://github.com/eclipse-tractusx/portal/blob/localdev-portal-iam-0.6.0/charts/localdev/README.md)) the processes worker needs multiple tries to run through entirely when executing the provisioning in centralidp and sharedidp after the invitation. If this happens, the worker executes all steps until the creation of service account in the sharedidp (for instance saidp1) at the first try and at the second try the worker fails to continue with the provisioning because the service account in sharedidp already exists.

### Expected Behavior

Even if the processes worker needs multiple tries to run through entirely, the process steps should have been logged so that the worker picks up from the latest logged step.

### Steps To Reproduce

Detected in version 2.0.0

### Additional information

After analysing further it was noticed that, while the **missing locking during the invite process** was for sure a necessary improvement, the root cause of the issue was that in the [local development environment](https://github.com/eclipse-tractusx/portal/blob/localdev-portal-iam-0.6.0/charts/localdev/README.md) **no valid encryptionKeys** were configured, but noticing that misconfiguration was cumbersome due to insufficient exception handing.
Therefore, the exception handing in case of non-valid encryptionKeys was improved https://github.com/eclipse-tractusx/portal-backend/pull/790

PR #790, opened 2024-06-13
fix(invite): adjust exceptionhandling for encryption

ntruchsess:
## Description

Exception-handling was added to the CryptoHelper to properly map the System-exceptions that are thrown by the Aes-classes to Configuration resp. ConflictExceptions

## Why

In case encryption fails due to invalid setup (or unsuitable data) the dotnet aes-classes throw exceptions that are derived from SystemException. If this happens during the execution of a process-step the execution is terminated without final save. (This implies the process-step remains in status TODO). Whenever a remote system-call was executed before the SystemException is being thrown that remote-call would be executed again when re-executing the process-step the next time. This would result in errors if the remote-system-call is not reentrant.

## Issue

#787

## Checklist

- [X] I have followed the [contributing guidelines](https://github.com/eclipse-tractusx/portal-assets/blob/main/docs/developer/Technical%20Documentation/Dev%20Process/How%20to%20contribute.md#commit-and-pr-guidelines)
- [X] I have added tests that prove my changes work
- [X] I have checked that new and existing tests pass locally with my changes


Phil91 (approved):
The changes to the cryptoHelper looks good. We should add a issue to align all other usages of the crypto helper to the new solution to make sure the error doesn't happen in other use cases as well.

evegufy (approved):

Issue #920, opened 2024-08-12
Identity Management | Add Multiple Users | Invite email didnt receive by User(s)


tfjanjua:
### Current Behavior
Add user account by Add Multiple Users
User(s) didnt receive invite emails when invited via Add Multiple Users.
Users has been created but invite email does not sent out to users.

### Expected Behavior
Invited users should receive welcome email along with one time password.

### Steps To Reproduce
- Login to portal as CX Admin
- Navigate to User management → Identity user management --> Add multiple users
- Download template
- Add/modify data with actual emails
- Upload back and hit Confirm

### Findings
Code to send email after creating users as bulk was not placed.

PR #922, opened 2024-08-12
fix(emailNotification): Add Multiple Users | Invite email didnt receive by User(s)

tfjanjua:
## Description

User(s) didnt receive invite emails when invited via Add Multiple Users. Users has been created but invite email does not sent out to users.

## Why

Recently bulk uploaded users were not getting email notifications

## Issue

[Ref: #920](https://github.com/eclipse-tractusx/portal-backend/issues/920)

## Checklist

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/UserUploadBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/UserUploadBusinessLogic.cs
@@ -23,6 +23,7 @@ using Org.Eclipse.TractusX.Portal.Backend.Framework.ErrorHandling;
 using Org.Eclipse.TractusX.Portal.Backend.Framework.ErrorHandling.Service;
 using Org.Eclipse.TractusX.Portal.Backend.Framework.IO;
 using Org.Eclipse.TractusX.Portal.Backend.Framework.Linq;
+using Org.Eclipse.TractusX.Portal.Backend.PortalBackend.DBAccess;
 using Org.Eclipse.TractusX.Portal.Backend.PortalBackend.DBAccess.Models;
 using Org.Eclipse.TractusX.Portal.Backend.PortalBackend.PortalEntities.Enums;
 using Org.Eclipse.TractusX.Portal.Backend.PortalBackend.PortalEntities.Identities;
@@ -41,6 +42,7 @@ public class UserUploadBusinessLogic : IUserUploadBusinessLogic
     private readonly UserSettings _settings;
     private readonly IIdentityData _identityData;
     private readonly IErrorMessageService _errorMessageService;
+    private readonly IPortalRepositories _portalRepositories;
 
     /// <summary>
     /// Constructor.
@@ -50,18 +52,21 @@ public class UserUploadBusinessLogic : IUserUploadBusinessLogic
     /// <param name="identityService">Access to the identity Service</param>
     /// <param name="errorMessageService">ErrorMessage Service</param>
     /// <param name="settings">Settings</param>
+    /// <param name="portalRepositories">Portal Repositories</param>
     public UserUploadBusinessLogic(
         IUserProvisioningService userProvisioningService,
         IMailingProcessCreation mailingProcessCreation,
         IIdentityService identityService,
         IErrorMessageService errorMessageService,
-        IOptions<UserSettings> settings)
+        IOptions<UserSettings> settings,
+        IPortalRepositories portalRepositories)
     {
         _userProvisioningService = userProvisioningService;
         _mailingProcessCreation = mailingProcessCreation;
         _identityData = identityService.IdentityData;
         _errorMessageService = errorMessageService;
         _settings = settings.Value;
+        _portalRepositories = portalRepositories;
     }
 
     public ValueTask<UserCreationStats> UploadOwnCompanyIdpUsersAsync(Guid identityProviderId, IFormFile document, CancellationToken cancellationToken)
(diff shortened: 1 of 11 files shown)


Phil91 (changes requested):
My suggestion would be to move the mail process creation into the call of `_userProvisioningServiceCreateOwnCompanyIdpUsersAsync` and adjust the usages accordingly.

tfjanjua:
> CreateOwnCompanyIdpUsersAsync

@Phil91 thanks for your suggestion, I see `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method and I found that we are calling this method from 8 different places (except test cases, otherwise its 24 places) and a few place has different email template and params implementation, wherever this method is been called. 

e.g.
in `UserBusinessLogic.cs` and `UserUploadBusinessLogic.cs` files, there are mostly 3 email templates have been used after calling the `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method from different places:
1. NewUserTemplate
2. NewUserPasswordTemplate
3. NewUserExternalIdpTemplate

and in `InvitationProcessService.cs` file, there are 2 different email templates have been used after calling the `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method:
1. RegistrationTemplate
2. PasswordForRegistrationTemplate

and in `RegistrationBusinessLogic.cs` file, there are 2 different email templates have been used after calling the `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method:
1. invite or inviteWithMessage
2. password

So I feel like, if I move the mail process creation for `UserBusinessLogic.cs` and `UserUploadBusinessLogic.cs` files into the call of `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` so, this would be adjustable because here we have `IsSharedIdp` bit to check and make decision upon email templates but I am wondering how the identification of email templates (params are more or less same) would work for `InvitationProcessService.cs` and `RegistrationBusinessLogic.cs` files? would it possible to add another param in `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` to check from where this method is being called and make decision for templates? or I am missing any already placed variable with different values?

tfjanjua:
> > > My suggestion would be to move the mail process creation into the call of `_userProvisioningServiceCreateOwnCompanyIdpUsersAsync` and adjust the usages accordingly.
> 
> @Phil91 thanks for your suggestion, I see `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method and I found that we are calling this method from 8 different places (except test cases, otherwise its 24 places) and a few place has different email template and params implementation, wherever this method is been called.
> 
> e.g. in `UserBusinessLogic.cs` and `UserUploadBusinessLogic.cs` files, there are mostly 3 email templates have been used after calling the `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method from different places:
> 
> 1. NewUserTemplate
> 2. NewUserPasswordTemplate
> 3. NewUserExternalIdpTemplate
> 
> and in `InvitationProcessService.cs` file, there are 2 different email templates have been used after calling the `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method:
> 
> 1. RegistrationTemplate
> 2. PasswordForRegistrationTemplate
> 
> and in `RegistrationBusinessLogic.cs` file, there are 2 different email templates have been used after calling the `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method:
> 
> 1. invite or inviteWithMessage
> 2. password
> 
> So I feel like, if I move the mail process creation for `UserBusinessLogic.cs` and `UserUploadBusinessLogic.cs` files into the call of `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` so, this would be adjustable because here we have `IsSharedIdp` bit to check and make decision upon email templates but I am wondering how the identification of email templates (params are more or less same) would work for `InvitationProcessService.cs` and `RegistrationBusinessLogic.cs` files? would it possible to add another param in `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` to check from where this method is being called and make decision for templates? or I am missing any already placed variable with different values?

Phil91:
> > My suggestion would be to move the mail process creation into the call of `_userProvisioningServiceCreateOwnCompanyIdpUsersAsync` and adjust the usages accordingly.
> 
> @Phil91 thanks for your suggestion, I see `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method and I found that we are calling this method from 8 different places (except test cases, otherwise its 24 places) and a few place has different email template and params implementation, wherever this method is been called.
> 
> e.g. in `UserBusinessLogic.cs` and `UserUploadBusinessLogic.cs` files, there are mostly 3 email templates have been used after calling the `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method from different places:
> 
> 1. NewUserTemplate
> 2. NewUserPasswordTemplate
> 3. NewUserExternalIdpTemplate
> 
> and in `InvitationProcessService.cs` file, there are 2 different email templates have been used after calling the `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method:
> 
> 1. RegistrationTemplate
> 2. PasswordForRegistrationTemplate
> 
> and in `RegistrationBusinessLogic.cs` file, there are 2 different email templates have been used after calling the `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method:
> 
> 1. invite or inviteWithMessage
> 2. password
> 
> So I feel like, if I move the mail process creation for `UserBusinessLogic.cs` and `UserUploadBusinessLogic.cs` files into the call of `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` so, this would be adjustable because here we have `IsSharedIdp` bit to check and make decision upon email templates but I am wondering how the identification of email templates (params are more or less same) would work for `InvitationProcessService.cs` and `RegistrationBusinessLogic.cs` files? would it possible to add another param in `_userProvisioningService.CreateOwnCompanyIdpUsersAsync` method to check from where this method is being called and make decision of email templates? or am I missing any already placed variable with different values?

@tfjanjua sorry for the late response, didn't saw the additional usages. I like the idea, instead of the parameter to define where the call came from I'd suggest to just add a parameter to pass an enumerable of the mail templates?

Phil91 (approved):

src/administration/Administration.Service/BusinessLogic/UserUploadBusinessLogic.cs
```diff
@@ -167,6 +172,52 @@ await GetUserRoleDatas(parsed.Roles, validRoleData, _identityData.CompanyId).Con
...
+        var displayName = await _userProvisioningService.GetIdentityProviderDisplayName(companyNameIdpAliasData.IdpAlias).ConfigureAwait(ConfigureAwaitOptions.None) ?? companyNameIdpAliasData.IdpAlias;
+
+        await foreach (var result in
+            _userProvisioningService
+                .CreateOwnCompanyIdpUsersAsync(
+                    companyNameIdpAliasData,
+                    userCreationInfos
+                        .Select(info =>
+                        {
+                            userCreationInfo = info;
```
ntruchsess, line 188: I'm not sure this might be causing threading issues. I need to check what the specification says about synchronization and visibility of changes done from an inner Task writing to some outer scoped address that is not defined as AtomicXXX

ntruchsess (changes requested):

src/administration/Administration.Service/BusinessLogic/UserUploadBusinessLogic.cs
ntruchsess: ok, I just notice this new method is a copy of the existing method CreateOwnCompanyIdpUsersWithEmailAsync with email-generation adjusted to the SharedIdp-usecase. I'll do some reusable refactoring to accomplish this in a less fragile way.

ntruchsess: I did refactor this adding a new onSuccess parameter that allows to pass a lambda that is going to be executed after each user-creation. Any db-changes applied within that lambda are being saved within the same transaction as the creation of the respective user now.
(I had to do a force-push after rebasing on latest main to resolve the conflicts)

ntruchsess (approved):

ntruchsess (dismissed):

Phil91 (approved):

ntruchsess (approved):

Phil91 (approved):

PR #951, opened 2024-08-26
fix(DID): removed authentication details from logs DIM response

dhiren-singh-007:
## Description

Removed the authentication details from logs DIM response and only logged the DIM document

## Why

It was exposing the client id and secret of Dim wallet

## Issue

NA

## Checklist


dhiren-singh-007:
For now i have removed the secrets from the logs but i think we can utilize the `SecretOperator` which is used for removing secret and password from logs.
It may require further regex to identify the ClientSecret and it might take sometime .

Phil91 (approved):

Issue #979, opened 2024-09-06
Incorrect mapping of Region field with BPDM model


dhiren-singh-007:
### Current Behavior
User is getting null in Region field when using BPN autofill in Registration page even though it have the data in BPDM pool 

### Expected Behavior
Region field should be populated with value if exist

### Steps To Reproduce
1. Login to the Portal Development as CX-admin
2. Invite a Business Partner
3. Accept invite as invited user
4. Start the registration
5. Add BPN number in the field “Search for your Company data by BPN“ , use any BPN which have the data in pool
6. Check the populated data
![image](https://github.com/user-attachments/assets/ed927893-f7fd-441f-9f45-f837adec76b4)

### RCA
When Portal is sending the data to BPDM it send the Region data in District field of BPDM request model.
But when it fetching back the data from BPDM then it maps this Region field from AdministrativeAreaLevel1.RegionCode.
While it should map it with District field itself.

PR #980, opened 2024-09-06
fix(registration): fix region mapping

dhiren-singh-007:
## Description

Changed the mapping of Region Field with District 

## Why

User is sending this data from FE which is sent via BE using District field to BPDM but when user try to fill it again from BPN search it becomes null as it is mapped with another field in BPDM

## Issue

#979 

## Checklist


Phil91 (dismissed):

Phil91:
@MaximilianHauer should this be included in 2.2.0 or in main?

Phil91 (approved):

ntruchsess (approved):

Issue #942, opened 2024-08-20
Registration | Company Name Change Not Reflecting Consistently


dhiren-singh-007:
### Current Behavior
When inviting a Business Partner in CX Admin, the initial Company Name (e.g., ABC) was specified. However, during the company registration process via the access link provided in the invite email, the Company Name was changed to (e.g., XYZ). Despite the change being reflected correctly on the Business Partner invite page within CX Admin, it does not update uniformly on the company user side, where the original Company Name (ABC) remains visible when accessed through the user icon.

Additionally, the Company Name remains unchanged on the Catena-X Home Page.

### Expected Behavior
The Company Name displayed on the company user side should reflect the latest change made during the registration process. In this case, it should display "XYZ" after the Business Partner updates it during registration.

### Steps To Reproduce

1. Log in to CX Admin and initiate an invite for a Business Partner.
2. Specify Company Name as "ABC" during the initial invite setup.
3. Send the invite email containing access details to the Business Partner.
4. Business Partner accesses the registration link from the email and changes the Company Name to "XYZ" during the registration process.
5. Verify on the CX Admin side:
6. Check the Business Partner invite page in CX Admin. Note that the Company Name displayed reflects the change to "XYZ".
7. Verify on the company user side:
8. Click on the user icon associated with the invited Business Partner.
9. Observe that the Company Name displayed remains "ABC", instead of updating to "XYZ" as expected.
10. Verify on Catena-X Home Page.

Example screenshots:

1. Company invited as Neha -Email Template

2. On submit registration company name changed to Müller as shown 
![image](https://github.com/user-attachments/assets/7f9ebb7f-6a30-468c-b163-90f275802a28)

3. Now when company go to select company name in login screen , Central idp is still showing old company name 
![image](https://github.com/user-attachments/assets/fb95bdf5-be54-478b-8540-a9b34a8a0943)

4. Now company admin logged and open User icon as shown , it is still showing old company name 
![image](https://github.com/user-attachments/assets/6fcb3d80-7735-44d7-853d-4a0bb0d06736)

RCA : Old Company name is coming from display name and Organization mapper attribute which is part of IDP of that specific company.

Phil91:
@dhiren-singh-007 in my opinion adding the calls to update the organisation for the users of the company and a call to update the org name on the shared idp should be fine, since they are only updates, if it fails while updating retriggering the api is just fine.

dhiren-singh-007:
adding @MaximilianHauer

[...]

PR #1010, opened 2024-09-19
fix(registration): fixed the company name change in IDP

dhiren-singh-007:
## Description

Added method to update display name and organisation attribute value in Identity provider in keycloak

## Why

This was causing mismatch issue in login screen for company name and also in the user profile menu after  login in portal

## Issue

#942 

## Checklist

### DIFF
--- a/src/provisioning/Provisioning.Library/ProvisioningManager.cs
+++ b/src/provisioning/Provisioning.Library/ProvisioningManager.cs
@@ -187,7 +187,14 @@ public partial class ProvisioningManager : IProvisioningManager
     {
         var mappers = await _centralIdp.GetIdentityProviderMappersAsync(_settings.CentralRealm, idpAlias).ConfigureAwait(ConfigureAwaitOptions.None);
         var organisationMapperId = mappers.FirstOrDefault(z => z.Name == $"{_settings.MappedCompanyAttribute}-mapper")?.Id ?? string.Empty;
-        await UpdateAttributeInIdentityProvider(idpAlias, organisationMapperId, _settings.MappedCompanyAttribute, companyName);
+        if (!string.IsNullOrEmpty(organisationMapperId))
+        {
+            await UpdateAttributeInIdentityProvider(idpAlias, organisationMapperId, _settings.MappedCompanyAttribute, companyName);
+        }
+        else
+        {
+            await CreateCentralIdentityProviderOrganisationMapperAsync(idpAlias, companyName);
+        }
     }
 
     private async Task UpdateAttributeInIdentityProvider(string alias, string mapperId, string attributeName, string value) => await _centralIdp.UpdateIdentityProviderMapperAsync(
--- a/src/provisioning/Provisioning.Library/ProvisioningManager.cs
+++ b/src/provisioning/Provisioning.Library/ProvisioningManager.cs
@@ -21,6 +21,7 @@ using Microsoft.Extensions.Options;
 using Org.Eclipse.TractusX.Portal.Backend.Keycloak.ErrorHandling;
 using Org.Eclipse.TractusX.Portal.Backend.Keycloak.Factory;
 using Org.Eclipse.TractusX.Portal.Backend.Keycloak.Library;
+using Org.Eclipse.TractusX.Portal.Backend.Keycloak.Library.Models.IdentityProviders;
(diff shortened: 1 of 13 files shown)


src/provisioning/Provisioning.Library/ProvisioningManager.cs
```diff
@@ -182,6 +183,30 @@ public async ValueTask UpdateSharedIdentityProviderAsync(string alias, string di
         await UpdateCentralIdentityProviderAsync(alias, identityProvider).ConfigureAwait(ConfigureAwaitOptions.None);
     }
 
+    public async ValueTask UpdateCentralIdentityProviderMapperAsync(string alias, string displayName)
+    {
+        var mappers = await _centralIdp.GetIdentityProviderMappersAsync(_settings.CentralRealm, alias).ConfigureAwait(ConfigureAwaitOptions.None);
+        var organisationMapperId = mappers.FirstOrDefault(z => z.Name == $"{_settings.MappedCompanyAttribute}-mapper")?.Id ?? string.Empty;
```
ntruchsess, line 189: from my understanding the update will fail if organisationMapperId is not a valid id but an empty string. If the mapper is not found the method should delegate to the existing CreateCentralIdentityProviderOrganisationMapperAsync

src/provisioning/Provisioning.Library/ProvisioningManager.cs
```diff
@@ -182,6 +183,30 @@ public async ValueTask UpdateSharedIdentityProviderAsync(string alias, string di
         await UpdateCentralIdentityProviderAsync(alias, identityProvider).ConfigureAwait(ConfigureAwaitOptions.None);
     }
 
+    public async ValueTask UpdateCentralIdentityProviderMapperAsync(string alias, string displayName)
```
ntruchsess, line 186: this method should be relocated to class IdentityProviderMapper right after method CreateCentralIdentityProviderOrganisationMapperAsync and the name should be changed to indicate it is specific for the organisation-mapper (like UpdateCentralIdentityProviderOrganisationMapperAsync(string alias, string organisationName)
(private UpdateAttributeInIdentityProvider should be relocated as well)

ntruchsess (changes requested):
see comments

src/provisioning/Provisioning.Library/ProvisioningManager.cs
dhiren-singh-007: @ntruchsess Thanks for reviewing it . Can you please tell me which `CreateCentralIdentityProviderOrganisationMapperAsync` should i use as it exist in two file i.e 
```
src/provisioning/Provisioning.Library/Extensions/IdentityProviderManager.cs
src/externalsystems/Provisioning.Library/IdpManagement.cs
```

dhiren-singh-007: Hi @ntruchsess , i started this change today and got some issues , can you please clarify .
Issue : If i move this method`UpdateCentralIdentityProviderMapperAsync ` in `src/externalsystems/Provisioning.Library/IdpManagement.cs` then i have to move its parent method `UpdateCompanyNameInIdentityProvider` from `src/provisioning/Provisioning.Library/Service/UserProvisioningService.cs` to `src/externalsystems/Provisioning.Library/IdpManagement.cs` because currently `externalsystems/Provisioning.Library`  is referencing the  `provisioning/Provisioning.Library` .

If its ok then i can move all these methods in `src/externalsystems/Provisioning.Library/IdpManagement.cs` then I would like to double check with you if i can reference this project in Registration.Service ?

dhiren-singh-007: I think i did mistake in putting the above comment as it is part of different comment so i moved it in bottom.
otherwise delegate thing is fixed

src/provisioning/Provisioning.Library/ProvisioningManager.cs
dhiren-singh-007: Regarding the moving the method I commented here https://github.com/eclipse-tractusx/portal-backend/pull/1010#issuecomment-2432133484
Renaming is fixed

ntruchsess (approved):

Phil91 (approved):

Issue #1070, opened 2024-10-14
Create a dedicated endpoint for the applicationRequests that enables a CX-Operator to access docs from other companies 


MaximilianHauer:
### Description

Currently the endpoint `GET: /api/registration/documents/{documentId}` only response the document for the user which has uploaded the document. This check should be adjusted:

add a check if the requesting user is in a company which has the role `OPERATOR` if this is the case the user should be able to download the document.

the endpoint should check if the company application is not in "confirmed" otherwise it should raise a 403 "Documents not accessible as onboarding process finished"

### Acceptance Criteria

- [ ] Try to call the endpoint with an cx-operator role - the document should be downloaded 
- [ ] Try to call the endpoint with a role that is not cx-operator - you should receive an error message that your not allowed to access this endpoint 
- [ ] If the company's application is in the "confirmed" status, the endpoint should proceed with the document retrieval.
- [ ] If the company's application is not in the "confirmed" status, the endpoint should return a 403 HTTP status code with the message "Documents not accessible as onboarding process finished".

### Additional Information
The logic to be adjusted is `DocumentRepository` -> `GetDocumentIdWithCompanyUserCheckAsync` and `RegistrationBusinessLogic` -> `GetDocumentContentAsync`

MaximilianHauer:
@gopalreddy-bmw  can you test this ?

gopalreddy-bmw:
@MaximilianHauer Today i will test and close it.

MaximilianHauer:
@gopalreddy-bmw  tested ?

PR #1075, opened 2024-10-15
feat(check): additional checks for OPERATOR and confirmed status added in document request endpoint

AnuragNagpure:
## Description

Currently the endpoint GET: /api/registration/documents/{documentId} only response the document for the user which has uploaded the document. This check should be adjusted with adding check of OPERATOR and company application status confirmed.

## Why

These additional check required for getting document to particular user satisfies the check condition.

## Issue

Link to Github issue.

#1070 

### DIFF
--- a/src/registration/Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
+++ b/src/registration/Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
@@ -173,7 +173,7 @@ public class RegistrationBusinessLogic(
             throw new ForbiddenException($"The user is not permitted to access document {documentId}.");
         }
 
-        if (!documentDetails.IsStatusConfirmed)
+        if (documentDetails.IsStatusConfirmed)
         {
             throw new ForbiddenException($"Documents not accessible as onboarding process finished {documentId}.");
         }
--- a/src/registration/Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
+++ b/src/registration/Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
@@ -168,9 +168,14 @@ public class RegistrationBusinessLogic(
             throw new NotFoundException($"document {documentId} does not exist.");
         }
 
-        if (!documentDetails.IsSameUser)
+        if (!documentDetails.IsSameUser && !documentDetails.IsRoleOperator)
         {
             throw new ForbiddenException($"The user is not permitted to access document {documentId}.");
+        }
+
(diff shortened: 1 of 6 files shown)


src/registration/Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -168,11 +168,16 @@ public async Task UploadDocumentAsync(Guid applicationId, IFormFile document, Do
             throw new NotFoundException($"document {documentId} does not exist.");
         }
 
-        if (!documentDetails.IsSameUser)
+        if (!documentDetails.IsSameUser && !documentDetails.IsRoleOperator)
         {
             throw new ForbiddenException($"The user is not permitted to access document {documentId}.");
         }
 
+        if (!documentDetails.IsStatusConfirmed)
```
Phil91, line 176: It should only throw errors if an application is in status confirmed. Because than the company is onboarded and the operator shouldn't be able to access the companies documents.

tests/registration/Registration.Service.Tests/BusinessLogic/RegistrationBusinessLogicTest.cs
```diff
@@ -2918,6 +2918,23 @@ public async Task GetDocumentAsync_WithWrongUser_ThrowsForbiddenException()
         ex.Message.Should().Be($"The user is not permitted to access document {documentId}.");
     }
 
+    [Fact]
+    public async Task GetDocumentAsync_WithConfirmedApplicationStatus_ThrowsForbiddenException()
```
Phil91, line 2922: The title says if status confirmed it should throw an error but currently the test setup is that confirmed is false, please adjust the setup and the business logic afterwards.

Phil91 (changes requested):
Please add unit tests for `GetDocumentIdWithCompanyUserCheckAsync`

tests/registration/Registration.Service.Tests/BusinessLogic/RegistrationBusinessLogicTest.cs
AnuragNagpure: Adjusted and setup changed

src/registration/Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
AnuragNagpure: changed

Phil91 (approved):

Issue #1079, opened 2024-10-18
[PORTAL - BPN] Get data from partner network pool by portal backend


tfjanjua:
### Description

We need to develop an API in portal-backend to get data from partner-network-pool instead of calling business-partner-pool service directly from the portal-frontend. This approach will make it consistent and secure, as we are already been calling business-partner-pool service from portal-backed to get legal-entity data by BPN.

API: https://business-partners-pool.example.com/v6/members/legal-entities/search?page=0&size=10 

### Acceptance Criteria

- [x] Develop a portal-backend API that facilitates secure access to the endpoint of business-partner-pool and respond data to portal-frontend.
- [x] The portal-backend API should be authenticated first and then route to the business-partner-pool service internally.
- [x] The API should handle errors gracefully and returns appropriate HTTP status codes.
- [x] Expected config change: https://github.com/eclipse-tractusx/portal/issues/457
- [x] Add respective test cases in portal-backend repo.
- [x] Changes should be communicated to the front-end to make appropriate changes over there as well.

### Additional Information

1. There is a BpnAccess.cs file already exists in backend repo which is already been calling one of the business-partner-pool service's APIs to fetch the legal entity by BPN so, we can use same file and add another method to fetch partner-network-pool data.

2. Portal-Frontend is already been checking `view_partner_network` role to show partner-network table on UI so, same role (view_partner_network) can be used for this new API of portal-backend and get data from business-partner-pool service.

3. Currently portal-frontend is been directly calling business-partner-pool API:
url: https://business-partners-pool.example.com/v6/members/legal-entities/search?page=0&size=10 
request data: `{ "bpnLs": [""], "legalName": "" }`
OR
request data: `{ "bpnLs": ["BPNL000000000065"], "legalName": "" }`

- So, portal-backend can provide API to portal-frontend in a same way, small change in API route is expected.

### Expected Result / Test Cases

Functionality and Test Cases should work as they are working right now because with this change, portal-frontend would be getting partner-network data from portal-backend instead of getting data directly from business-partner-pool and portal-backend would be calling business-partner-pool service internally and responding data to frontend.

### Corresponding Configuration Ticket

https://github.com/eclipse-tractusx/portal/issues/457

MaximilianHauer:
@ntruchsess / @Phil91 / @evegufy  as i remember we discussed on this topic in the open meeting. 
[...]

PR #1082, opened 2024-10-21
feat(partnerNetwork): Get data from partner network pool by portal backend (#61)

tfjanjua:
## Description

Developed an API in portal-backend to get data from partner-network-pool instead of calling business-partner-pool service directly from the portal-frontend.

## Why

This approach will make it consistent and secure, as we are already been calling business-partner-pool service from portal-backed to get legal-entity data by BPN.

## Issue

Ref: #1079

## Corresponding Configuration PR

https://github.com/eclipse-tractusx/portal/pull/461

## Checklist

### DIFF
--- a/docs/api/administration-service.yaml
+++ b/docs/api/administration-service.yaml
@@ -3182,6 +3182,52 @@ paths:
           description: Internal Server Error
         '401':
           description: The User is unauthorized
+  /api/administration/PartnerNetwork/legalEntities/search:
+    post:
+      tags:
+        - PartnerNetwork
+      summary: 'Gets partner network data from BPN Pool (Authorization required - Roles: view_partner_network)'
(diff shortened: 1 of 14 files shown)


MaximilianHauer:
before merging/approving please look into the comment of the issue https://github.com/eclipse-tractusx/portal-backend/issues/1079

ntruchsess (commented):
This is not doing anything else than mapping BpdmPartnerNetworkData to PartnerNetworkResponse. No access to portal db is done, this implements more or less a reverse proxy. From my point of view @oyo 's comment https://github.com/eclipse-tractusx/portal-backend/issues/1079#issuecomment-2467532295 is valid and the implementation of a reverse-proxy should be considered instead.

tfjanjua:
for the time being, we are not contributing this change to the reference implementation, this change would only be part of Cofinity-X. I will close the corresponding Tickets as well.

Issue #1085, opened 2024-10-21
Only provider should be able to update the connector url 


dhiren-singh-007:
### Current Behavior
Currently in portal backend there is a validation that only host can update the connector url which restricts the provider to update the connector url.

### Expected Behavior
Actually provider is the one who provides the connector so only provider should be able to update the connector url.

### Steps To Reproduce
Provider : ABC
Host : XYZ
1. Create a managed connector for company "XYZ"
2. Update the connector url which is created for "Host Company" , it will throw and error "Company ABC is not the connectors host company"
3. Now login form XYZ company and update the managed connector, now it will allow the update.

ntruchsess:
@dhiren-singh-007 @MaximilianHauer : I think the behavior should depend on the ConnectorType:
* if it's ConnectorTypeId.CONNECTOR_AS_A_SERVICE the provider should be able to update
* if it's ConnectorTypeId.COMPANY_CONNECTOR the host should be able to update

dhiren-singh-007:
@ntruchsess : I think if COMPANY_CONNECTOR type can have a different provider and host then i think it make sense to allow host to update the connector.
But if COMPANY_CONNECTOR will always have provider and host as same id then i think provider validation should be enough.
cc @MaximilianHauer

ntruchsess:
@dhiren-singh-007 : you are right - ConnectorTypeId.COMPANY_CONNECTOR typed connectors are being created with host and provider being set to the same id ( https://github.com/eclipse-tractusx/portal-backend/blob/3a07d788937dc905d9939307f571fa000b535145/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs#L115 ). It's therefore sufficient (and also safe) to just check the providerid.
cc @MaximilianHauer

gopalreddy-bmw:
Hi @MaximilianHauer 

I tested this issue it's working fine. Can you close this ticket.

Attaching screenshots for references 
<img width="956" alt="Image" src="https://github.com/user-attachments/assets/08723c83-841c-418d-ba45-b25404c74e7f" />
<img width="959" alt="Image" src="https://github.com/user-attachments/assets/e39d7914-8567-4b97-95fd-2ba0a520920d" />

PR #1086, opened 2024-10-21
fix(connector): only provider should update connector

dhiren-singh-007:
## Description

Updated the validation on update connector api so that only provider can update the connector

## Why

Currently it was not possible for provider to update the connector url which provided by themself only

## Issue

#1085 

## Checklist

### DIFF
--- a/src/portalbackend/PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
+++ b/src/portalbackend/PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
@@ -192,7 +192,7 @@ public class ConnectorsRepository(PortalDbContext dbContext) : IConnectorsReposi
             .Select(c => new ConnectorUpdateInformation(
                 c.StatusId,
                 c.TypeId,
-                c.HostId == companyId,
+                c.ProviderId == companyId,
                 c.ConnectorUrl,
                 c.Provider!.BusinessPartnerNumber,
                 c.SelfDescriptionDocumentId,
--- a/tests/portalbackend/PortalBackend.DBAccess.Tests/ConnectorRepositoryTests.cs
+++ b/tests/portalbackend/PortalBackend.DBAccess.Tests/ConnectorRepositoryTests.cs
@@ -432,7 +432,6 @@ public class ConnectorRepositoryTests : IAssemblyFixture<TestDbFixture>
         // Assert
         result.Should().NotBeNull();
         result!.IsProviderCompany.Should().BeTrue();
-
     }
 
     [Fact]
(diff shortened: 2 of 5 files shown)


Phil91 (approved):

src/portalbackend/PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
```diff
@@ -190,7 +190,7 @@ public Connector AttachAndModifyConnector(Guid connectorId, Action<Connector>? i
             .Select(c => new ConnectorUpdateInformation(
                 c.StatusId,
                 c.TypeId,
-                c.HostId == companyId,
```
Phil91, line 193: Since there seems to be no unit tests, please add a unit test to check that the correct value is set.

dhiren-singh-007: added

ntruchsess (changes requested):
please see my comment in https://github.com/eclipse-tractusx/portal-backend/issues/1085#issuecomment-2476753684

tests/portalbackend/PortalBackend.DBAccess.Tests/ConnectorRepositoryTests.cs
```diff
@@ -442,7 +442,10 @@ public async Task GetConnectorUpdateInformation_ReturnIsProviderTrue()
 
         // Assert
         result.Should().NotBeNull();
-        result?.IsProviderCompany.Should().BeTrue();
+        if (result != null)
+        {
+            result.IsProviderCompany.Should().BeTrue();
+        }
```
ntruchsess, line 448: the null check is redundant as there's a NotBeNull assert right above. As this is a unit-test the use of '!' operator is more than justified in such cases.
```suggestion
        result!.IsProviderCompany.Should().BeTrue();
```

tests/portalbackend/PortalBackend.DBAccess.Tests/ConnectorRepositoryTests.cs
```diff
@@ -456,7 +459,10 @@ public async Task GetConnectorUpdateInformation_ReturnIsProviderFalse()
 
         // Assert
         result.Should().NotBeNull();
-        result?.IsProviderCompany.Should().BeFalse();
+        if (result != null)
+        {
+            result.IsProviderCompany.Should().BeFalse();
+        }
```
ntruchsess, line 465:
```suggestion
        result.!IsProviderCompany.Should().BeFalse();
```

ntruchsess (changes requested):

tests/portalbackend/PortalBackend.DBAccess.Tests/ConnectorRepositoryTests.cs
dhiren-singh-007: Thanks @ntruchsess ,Fixed

Phil91 (approved):
lgtm

ntruchsess (approved):

Issue #549, opened 2024-03-03
Administration Service - Connector Controller change connector URL


jjeroch:
### Summary
Recheck and enhancement of the current existing PUT endpoint(s) for connector controller

### Details

**Endpoint:** `/api/administration/Connectors/{connectorId}/connectorUrl`

**Method:** `PUT`

**Needed changes:**
- Implementation of following backend logic:
  - with the endpoint trigger, the backend logic will set the current connector assigned self-description to "INACTIVE" - no matter which status the document has currently (retrieve via the table `portal.connectors` the linked self-description document from the attribute connectors.self_description_document_id
  - open the table `portal.documents`, search for the connectors.self_description_document_id inside the field `id` and set the status to "INACTIVE" 
  - now trigger the SD factory endpoint (same as triggered for the POST `/api/administration/Connectors` - functional details: https://github.com/eclipse-tractusx/portal-assets/blob/main/docs/developer/02.%20Technical%20Integration/01.%20Connector%20Registration/05.%20Self%20Description.md) and submit the relevant data

AnuragNagpure:
@jjeroch the description is yet to update as we discussed in parking lot on that day. Can you please check once.

jjeroch:
@AnuragNagpure  please check if the updated ticket description is enough or needs more information

MaximilianHauer:
@AnuragNagpure  do you need support on this ? did you see @jjeroch  comment ?

AnuragNagpure:
> @AnuragNagpure do you need support on this ? did you see @jjeroch comment ?

no currently understood will do accordingly

PR #1094, opened 2024-10-22
feat(connector): connector controller change url added

AnuragNagpure:
## Description

in endpoint api/administration/Connectors/{connectorId}/connectorUrl
open the table portal.documents, search for the connectors.self_description_document_id inside the field id and set the status to "INACTIVE"
now trigger the SD factory endpoint 

## Why

The backend logic will set the current connector assigned self-description to "INACTIVE" - no matter which status the document has currently (retrieve via the table portal.connectors the linked self-description document from the attribute connectors.self_description_document_id.

## Issue

#549 

## Checklist

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
@@ -437,15 +437,20 @@ public class ConnectorsBusinessLogic(
 
         if (connector.SelfDescriptionDocumentId != Guid.Empty)
         {
-            documentRepository.AttachAndModifyDocument((Guid)connector.SelfDescriptionDocumentId!, null, doc =>
+            documentRepository.AttachAndModifyDocument(connector.SelfDescriptionDocumentId.Value, null, doc =>
             {
                 doc.DocumentStatusId = DocumentStatusId.INACTIVE;
             });
         }
 
+        if (connector.SelfDescriptionCompanyDocumentId is null)
+        {
+            throw ConflictException.Create(AdministrationConnectorErrors.CONNECTOR_CONFLICT_NO_DESCRIPTION, [new("connectorId", connectorId.ToString())]);
+        }
+
         var selfDescriptionDocumentUrl = $"{_settings.SelfDescriptionDocumentUrl}/{connector.SelfDescriptionCompanyDocumentId}";
         await sdFactoryBusinessLogic
-            .RegisterConnectorAsync(connectorId, selfDescriptionDocumentUrl, connector.Bpn!, cancellationToken)
+            .RegisterConnectorAsync(connectorId, selfDescriptionDocumentUrl, bpn, cancellationToken)
             .ConfigureAwait(ConfigureAwaitOptions.None);
 
         await portalRepositories.SaveAsync().ConfigureAwait(ConfigureAwaitOptions.None);
--- a/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
@@ -384,16 +384,18 @@ public class ConnectorsBusinessLogic(
     }
 
     /// <inheritdoc />
-    public Task UpdateConnectorUrl(Guid connectorId, ConnectorUpdateRequest data)
+    public Task UpdateConnectorUrl(Guid connectorId, ConnectorUpdateRequest data, CancellationToken cancellationToken)
     {
         data.ConnectorUrl.EnsureValidHttpUrl(() => nameof(data.ConnectorUrl));
-        return UpdateConnectorUrlInternal(connectorId, data);
+        return UpdateConnectorUrlInternal(connectorId, data, cancellationToken);
     }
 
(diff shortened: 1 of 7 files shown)


src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -433,6 +435,19 @@ private async Task UpdateConnectorUrlInternal(Guid connectorId, ConnectorUpdateR
             con.ConnectorUrl = data.ConnectorUrl;
         });
 
+        if (connector.SelfDescriptionDocumentId != Guid.Empty)
+        {
+            documentRepository.AttachAndModifyDocument((Guid)connector.SelfDescriptionDocumentId!, null, doc =>
```
Phil91, line 440: Please change to .Value for easier readability

```suggestion
            documentRepository.AttachAndModifyDocument(connector.SelfDescriptionDocumentId.Value, null, doc =>
```

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -433,6 +435,19 @@ private async Task UpdateConnectorUrlInternal(Guid connectorId, ConnectorUpdateR
...
+            {
+                doc.DocumentStatusId = DocumentStatusId.INACTIVE;
+            });
+        }
+
+        var selfDescriptionDocumentUrl = $"{_settings.SelfDescriptionDocumentUrl}/{connector.SelfDescriptionCompanyDocumentId}";
```
Phil91, line 446: Since the `connector.SelfDescriptionCompanyDocumentId` is nullable this should be checked... otherwise the callback from the sdFactory will always fail... If it is null you can throw a conflict exception

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -433,6 +435,19 @@ private async Task UpdateConnectorUrlInternal(Guid connectorId, ConnectorUpdateR
...
+        await sdFactoryBusinessLogic
+            .RegisterConnectorAsync(connectorId, selfDescriptionDocumentUrl, connector.Bpn!, cancellationToken)
```
Phil91, line 448: The bpn should be checked for null as well.

tests/administration/Administration.Service.Tests/BusinessLogic/ConnectorsBusinessLogicTests.cs
```diff
@@ -1112,14 +1117,23 @@ public async Task UpdateConnectorUrl_WithValidData_CallsExpected()
...
-        await _logic.UpdateConnectorUrl(connectorId, new ConnectorUpdateRequest("https://new.de"));
+        await _logic.UpdateConnectorUrl(connectorId, new ConnectorUpdateRequest("https://new.de"), CancellationToken.None);
 
         // Assert
         A.CallTo(() => _portalRepositories.SaveAsync()).MustHaveHappenedOnceExactly();
         A.CallTo(() => _connectorsRepository.AttachAndModifyConnector(connectorId, null, A<Action<Connector>>._)).MustHaveHappenedOnceExactly();
+        A.CallTo(() => _documentRepository.AttachAndModifyDocument((Guid)data.SelfDescriptionDocumentId!, null, A<Action<Document>>._)).MustHaveHappenedOnceExactly();
         connector.ConnectorUrl.Should().Be("https://new.de");
+        document.DocumentStatusId.Should().Be(DocumentStatusId.INACTIVE);
     }
```
Phil91, line 1137: Please also check that the sdFactory was called as expected

tests/administration/Administration.Service.Tests/BusinessLogic/ConnectorsBusinessLogicTests.cs
Phil91, line 1: After you added the null checks for bpn and CompanySelfDescriptionDocument please provide unit tests as well

Phil91 (changes requested):

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
AnuragNagpure: Changed

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
AnuragNagpure: Added the check

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
AnuragNagpure: there is already check for bpn with error msg

tests/administration/Administration.Service.Tests/BusinessLogic/ConnectorsBusinessLogicTests.cs
AnuragNagpure: added

Phil91 (approved):

Issue #1111, opened 2024-10-23
Process Worker | Set Catena-X Membership in BPDM


tfjanjua:
### Description

As Portal, we would like to maintain a list of all companies that were onboarded to the Catena-X network. This membership information is meant to be stored in the BPDM Pool. We therefore need to update the `isCatenaXMember` flag for a BPNL once the company has been onboarded. 

The BPDM Pool offers an endpoint to update the `isCatenaXMember` flag via an endpoint. Example to update the membership information for a given BPNL. Please note that the path needs to be adapted to the specific environment.

```
curl -X 'PUT' \
  'http://bpdm-example.com/v6/cx-memberships' \
  -H 'accept: */*' \
  -H 'Content-Type: application/json' \
  -d '{
  "memberships": [
    {
      "bpnL": "BPNL000000000001",
      "isCatenaXMember": true
    }
  ]
}'

```

### Corresponding Configuration Ticket

https://github.com/eclipse-tractusx/portal/issues/471

### Corresponding BPDM PR

PR: https://github.com/eclipse-tractusx/bpdm/pull/1072

### Acceptance Criteria

- [x] New ApplicationChecklist process step `SET_CX_MEMBERSHIP_IN_BPDM` need be added for setting the Catena-X membership in BPDM.
- [x] This new ApplicationChecklist process step would be calling above mentioned PUT api call and setting the `isCatenaXMember` status.
- [x] This new ApplicationChecklist process step would have re-trigger option as well.
- [x] This new ApplicationChecklist process step need to be called after `MEMBERSHIP_CREDENTIAL` step but this new step could also be called as part of `APPLICATION_ACTIVATION` type and called after the `SET_MEMBERSHIP` step.

### Test Cases

Test Case 1: Verify `SET_CX_MEMBERSHIP_IN_BPDM` step as part of process worker

- Navigate to https://portal.example.com/invite and invite Business Partner.
- Complete the registration of newly invited Business Partner.
- Navigate to https://portal.example.com/applicationRequests and let the process worker work.
- Verify the all steps to be completed by Process Worker.

Test Case 2: Verify `SET_CX_MEMBERSHIP_IN_BPDM` step as part of re-trigger process.

- Navigate to https://portal.example.com/invite and invite Business Partner.
- Complete the registration of newly invited Business Partner.
- Navigate to https://portal.example.com/applicationRequests and let the process worker work.
[...]

PR #1118, opened 2024-10-25
feat(CxMembership): Set Catena-X Membership in BPDM

tfjanjua:
## Description

Set Catena-X Membership in BPDM.

## Why

As Portal, we would like to maintain a list of all companies that were onboarded to the Catena-X network. This membership information is meant to be stored in the BPDM Pool. We therefore need to update the `isCatenaXMember` flag for a BPNL once the company has been onboarded.

## Issue

Ref: #1111

Config PR: https://github.com/eclipse-tractusx/portal/pull/472

BPDM PR: https://github.com/eclipse-tractusx/bpdm/pull/1072

## Checklist


src/externalsystems/Bpdm.Library/BpdmService.cs
```diff
@@ -176,4 +177,23 @@ public async Task<BpdmSharingState> GetSharingState(Guid applicationId, Cancella
...
     }
+
+    /// <inheritdoc />
+    public async Task<bool> SetCxMembership(string businessPartnerNumber, CancellationToken cancellationToken)
+    {
+        using var httpClient = await _tokenService.GetAuthorizedClient<BpdmService>(_settings, cancellationToken).ConfigureAwait(ConfigureAwaitOptions.None);
+
+        // Same User (which we have for BPDM service) can be used to call Business Partner Pool
+        // But BaseAddress of Business Partner Pool is different as its deployed on another server.
+        httpClient.BaseAddress = new Uri(_settings.BusinessPartnerPoolBaseAddress);
```
ntruchsess, line 188: (while technically working) this is not how a httpClientFactory (which is used in tokenService) should be used. The BusinesCode should not have to deal with the baseAddress.

ntruchsess (changes requested):
I did fix a few findings and force-pushed to your repo after rebasing on latest main. Please pull he changes before proceeding. (And also have a look on what I changed to avoid such findings in the future).
PR currently cannot be merged as the configuration in portal-repo does not match. Also we need to come up with a better approach to handle multiple bpdm deployments without hardcoding the fact that this particular call is different into the service-method.

tfjanjua:
> > @tfjanjua : I did fix a few findings and force-pushed to your repo after rebasing on latest main. Please pull he changes before proceeding. (And also have a look on what I changed to avoid such findings in the future). PR currently cannot be merged as the configuration in portal-repo does not match. Also we need to come up with a better approach to handle multiple bpdm deployments without hardcoding the fact that this particular call is different into the service-method.
> 
> Thank you @ntruchsess , providing your feedback.
> 
> I've looked at your commit and found some formatting and addition of license related information (especially in migration files).
> 
> I'll have a look at configuration PR.

@ntruchsess , I have added my comment in the corresponding config PR: https://github.com/eclipse-tractusx/portal/pull/472#discussion_r1854239204

You may resolve the conversation if the provided explanations make sense to you as well.

Thanks

src/externalsystems/Bpdm.Library/BpdmService.cs
tfjanjua: @ntruchsess I totally agree to not assign value to `BaseAddress` param at library when its already managed in common token service but here we have a situation that we want to call this newly introduced method `SetCxMembership` from background process and not from the controller (because there we have a possibility to take user's token use same token to call Business Partner Pool api route) and same user (which has been used to call BPDM services) can be used to call Business Partner Pool service so, thats why I introduced another kind of baseAddress `BusinessPartnerPoolBaseAddress` within the same BPDM configs and override the value of BaseAddress of BPDM config before calling the tokenService so, I could use same user (which has been used to call BPDM services) and could get the token to call this Business Partner Pool service.

It would be nice if we could separate out BpnAccess service from BPDM library because I think then we can avoid assigning or overriding the value of BaseAddress because then we would have opportunity to have proper user (with baseAddress) to call Business Partner Pool services (no matter from controller or background process).

evegufy:
@tfjanjua could you please resolve the conflict?
@ntruchsess @Phil91 could you please review?

tfjanjua:
> @tfjanjua : I did fix a few findings and force-pushed to your repo after rebasing on latest main. Please pull he changes before proceeding. (And also have a look on what I changed to avoid such findings in the future). PR currently cannot be merged as the configuration in portal-repo does not match. Also we need to come up with a better approach to handle multiple bpdm deployments without hardcoding the fact that this particular call is different into the service-method.

Hi @ntruchsess , I forgot to take pull before resolving the conflicts and lost your finding's fix but I have tried to find your findings (like removing bom from utf-8 encoding and adding a proper license info especially on newly created files) and tried to fix them, please let me know if I have missed any point.

tfjanjua:
@evegufy | @Phil91 | @ntruchsess 
Please let me know if there is any other finding so, we can resolve them and merge this PR as soon as possible. Thanks

evegufy:
@tfjanjua could you please solve the conflict?
@Phil91 @ntruchsess could you please review? The PR is open since quite some time now.

src/externalsystems/Bpdm.Library/BpdmService.cs
Phil91: @tfjanjua I agree with @ntruchsess instead of changing the url in the business logic we should adjust the `BpnAccess` from using the token of the user to using a technical user. Therefore we would need to add a technical user with the needed roles (we should discuss with @nicoprow which roles would be needed to set it up properly) after that we can use the bpnAccess for the web services as well as the process worker.

Phil91 (changes requested):

ntruchsess:
@Phil91 : regarding your comment https://github.com/eclipse-tractusx/portal-backend/pull/1118#discussion_r1930683096
from what I can see the new code is not using the current user's token - it's using whatever is configured as technical user in the BpdmServiceSettings,

ntruchsess (approved):
@Phil91 : regarding your comment https://github.com/eclipse-tractusx/portal-backend/pull/1118#discussion_r1930683096
from what I can see the new code is not using the current user's token - it's using whatever is configured as technical user in the BpdmServiceSettings, So from my point of view this part is fine.
@tfjanjua + @Phil91 : I did refactor the TokenService to support more httpclient-names than just the classname per set of technical-user-credentials. This allows to set the baseUrl for bpdm-pool-service in the httpClient-configuration (removing is from the businessLogic). To solve the conflicts I did also rebase on main and recreated the migration doing a force-push to your repository. Please pull those changes before doing any further changes. (From my point of view no changes should be required though)

Phil91 (approved):

Issue #1114, opened 2024-10-24
Process worker Crashing pods on System Exception


dhiren-singh-007:
### Current Behavior
Whole Processworker fails if there is any SystemException , for eg If there is no email smtp is configured then it crashed the whole process worker which impacts the processing of the other processes. 

### Expected Behavior
Exception within some process should not impact other processes. All processes should run independently.

### Steps To Reproduce
1. Remove Email SMTP address from environment 
2. Invite parner or Subscribe to apps or anything which can create process
3. Check logs of process worker .Process will be throw SocketExcpetion which will be crashing the pod.
4. Other process will not be executed until email address is set up

PR #1119, opened 2024-10-25
fix(process worker): fix the crashing of pod

dhiren-singh-007:
## Description

Handled the system exception and added the Critical/Fatal log type so that we can monitor the logs for criticality.

## Why

It was causing issues for other processes as well .
Eg if a OSP customer sets the wrong callback address or if it is down then it returns the referenced class of SystemException and then it crashed the pod , which stops all other processes.

## Issue

#1114 

## Checklist

### DIFF
--- a/src/framework/Framework.Async/Directory.Build.props
+++ b/src/framework/Framework.Async/Directory.Build.props
@@ -19,7 +19,7 @@
 
 <Project>
   <PropertyGroup>
-    <VersionPrefix>2.12.0</VersionPrefix>
+    <VersionPrefix>2.13.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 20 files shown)


Phil91 (changes requested):
Please revert the changes in the framework dir since nothing changed in the framework logic

ntruchsess (approved):

ntruchsess:
@dhiren-singh-007 : I think it makes sense to force a restart of the pod when it runs out of memory - everything else should be logged as critical. I also changed the way SocketException and IOException are handled within the remote api calls. Those wouln't throw this kind of systemexception but ServiceException flagged as RecoverOptions.NETWORK (being mapped to RecoverOptions.INFRASTRUCTURE which is most external system calls treat as recoverable skipping the process-step to be retried later). Nevertheless we might want to evaluate whether this is the right strategy for all external system calls - e.g. when calling a url that is provided by the customer it might be better to set the respective process-step to ERROR in this case.

Phil91 (approved):

Issue #1089, opened 2024-10-21
EDC discovery is returning wrong connector endpoints


dhiren-singh-007:
API: `GET : /api/administration/connectors/discovery`
Request : `["BPN"]`

### Current Behavior
EDC discovery api is returning the provider connector under provider BPN

### Expected Behavior
EDC discovery api should return the Company connector and Managed connector under Host BPN
For eg : 
Provider company "ABC" created the Managed connector for "XYZ" company with connector url `https://connectorurl`
Now when we call the EDC discovery api it returns the response like below
```json
{
 [
   "bpn":"ABC"
   "connectorEndpoint":
   [
   "https://connectorurl"
   ]
 ]
}
```
while response should be like 
```json
{
 [
   "bpn":"XYZ"
   "connectorEndpoint":
   [
   "https://connectorurl"
   ]
 ]
}
```

Assuming "ABC" and "XYZ" are the BPN of company
### Steps To Reproduce
1. Create a managed connector for host company XYZ
2. Now call this api `/api/administration/connectors/discovery` with request `["XYZ"]` , it will not return the newly connected connector
3.  Now call this api again `/api/administration/connectors/discovery` with request `["ABC"]` , it will return the newly connected connector under this BPN.

dhiren-singh-007:
@MaximilianHauer @ntruchsess @Phil91 : As we discussed i created this issue for edc discovery api.
Please review and share your thoughts and then i will raise the PR accordingly.

PR #1120, opened 2024-10-28
fix(connector): fixed the edc discovery endpoint

dhiren-singh-007:
## Description

Fixed the EDC discovery endpoint by filtering an grouping it by Host 

## Why

Earlier its was filtering by provider which was showing managed connector under provider BPN which is incorrect.

## Issue

#1089 

## Checklist


dhiren-singh-007:
@Phil91 : Can i update the Dependency as a blank file because in artifacts it is blank 
https://github.com/eclipse-tractusx/portal-backend/actions/runs/11551164069?pr=1120
Please help

Phil91:
> @Phil91 : Can i update the Dependency as a blank file because in artifacts it is blank https://github.com/eclipse-tractusx/portal-backend/actions/runs/11551164069?pr=1120 Please help

@dhiren-singh-007 sometimes the workflow fails because the dependencies aren't build correctly. I'll restart the workflow again

Phil91 (changes requested):
Since this change didn't affect the unit tests at all, could you please add a test case that would actually fail when using Provider instead of Host.

Phil91 (approved):

dhiren-singh-007:
@Phil91 I thought that two reviewers which i selected will be treated as optional if one approves , so removed the other one.

Issue #1217, opened 2024-12-27
Compliance GX Tagus Release V2 Api | Integrate the Gaia-X Clearinghouse Interface


tfjanjua:
### Description

Integrate the V2 api from Gaia-X Tagus release in company onboarding and SD document creation. 

Gaia-X has released the V2 api in Tagus release which have changes in request and response interfaces.
So portal backend have to adapt these new changes.

This new version have more flexibility from Gaia-X perspective and Portal BE must switch to new api for correct data validation for companies onboarding and Sd document creation.

Currently Portal BE is using V1 api from Gaia-X

### Acceptance Criteria

- [x] Portal backend would need to analyse the V2 api from Gaia-X and then integrate the changes in backend code for new interfaces.
- [x] Company onboarding should be successful by using V2 api from clearing house.

### Test Cases

**Test Case 1:**  Verify the Company Onboarding Clearing house with valid VAT_ID 
  
Test Steps: 
* Invite a Business Partner 
* Accept invite as invited Customer 
* Start with registration process 
* Fill out all registration details with valid & real VAT_ID 
* Submit registration data 
* Observe Company Onboarding process 

Expected Results: 
Clearing house should process the data successfully. 
  
 
**Test Case 2:** Verify the Company Onboarding Clearing house with invalid VAT_ID 

Test Steps: 
* Invite a Business Partner 
* Accept invite as invited Customer 
* Start with registration process 
* Fill out all registration details with invalid VAT_ID 
* Submit registration data 
* Observe Company Onboarding process 

Expected Results: 
Clearing house should not process the data successfully. 
  

**Test Case 3:** Company Onboarding with valid Commercial Registration identifier. 

Test Steps: 
* Invite a Business Partner 
* Accept invite as invited Customer 
* Start with registration process 
* Fill out all registration details with valid & real Commercial registration number 
* Submit registration data 
* Observe Company Onboarding process 

Expected Results: 
Clearing house should process the data successfully. 
 

**Test Case 4:** Company Onboarding with invalid Commercial Registration identifier 

Test Steps: 
* Invite a Business Partner 
* Accept invite as invited Customer 
* Start with registration process 
* Fill out all registration details with invalid/fake Commercial registration number 
* Submit registration data 
* Observe Company Onboarding process 

Expected Results: 
Clearing house should not process the data successfully. 
 

**Test Case 5:** Company Onboarding with valid LEI code 

Test Steps: 
* Invite a Business Partner 
[...]

PR #1220, opened 2025-01-03
feat(clearingHouse): Integrate the Tagus Release V2 Api

tfjanjua:
## Description

Integration of the v2 api from Gaia-X Tagus release in company onboarding clearing house.

## Why

v1 will be retired as of March 2025. Any API v1 calls will fail with a 400 HTTP Status Code.

## Issue

[Ref: 1217](https://github.com/eclipse-tractusx/portal-backend/issues/1217)

## Checklist


src/portalbackend/PortalBackend.DBAccess/Repositories/ApplicationRepository.cs
```diff
@@ -404,16 +405,19 @@ public IQueryable<CompanyApplication> GetAllCompanyApplicationsDetailsQuery(stri
...
-                    ca.Address.Zipcode,
-                    ca.Address.Country!.CountryLongNames.Where(cln => cln.ShortName == "en").Select(cln => cln.LongName).SingleOrDefault(),
-                    ca.Address.CountryAlpha2Code),
-                ca.CompanyIdentifiers.Select(ci => new UniqueIdData(ci.UniqueIdentifier!.Label, ci.Value))))
+                    new LegalAddress(
+                        ca.Address!.CountryAlpha2Code,
+                        string.Format("{0}-{1}", ca.Address!.CountryAlpha2Code, ca.Address!.Region),
+                        ca.Address.City,
+                        ca.Address.Zipcode,
+                        !string.IsNullOrEmpty(ca.Address.Streetnumber) ? string.Format("{0} {1}", ca.Address.Streetname, ca.Address.Streetnumber) : ca.Address.Streetname
```
Phil91, line 416: could you please switch for easier readability

```suggestion
                        string.IsNullOrEmpty(ca.Address.Streetnumber) ? ca.Address.Streetname : string.Format("{0} {1}", ca.Address.Streetname, ca.Address.Streetnumber) 
```

src/portalbackend/PortalBackend.DBAccess/Repositories/ApplicationRepository.cs
```diff
@@ -404,16 +405,19 @@ public IQueryable<CompanyApplication> GetAllCompanyApplicationsDetailsQuery(stri
             .Select(ca => new { ca.ApplicationStatusId, ca.Company, ca.Company!.Address, ca.Company.CompanyIdentifiers })
             .Select(ca => new ClearinghouseData(
                 ca.ApplicationStatusId,
-                new ParticipantDetails(
+                ca.Company!.BusinessPartnerNumber!,
```
Phil91, line 408: Since the address of a company could in theory be null the selection of the fields will probably fail, since they are not nullable. we should think about adding a check for address != null

src/portalbackend/PortalBackend.DBAccess/Repositories/ApplicationRepository.cs
```diff
@@ -404,16 +405,19 @@ public IQueryable<CompanyApplication> GetAllCompanyApplicationsDetailsQuery(stri
...
-                    ca.Address.Streetname,
-                    ca.Company.BusinessPartnerNumber,
-                    ca.Address.Region,
-                    ca.Address.Zipcode,
-                    ca.Address.Country!.CountryLongNames.Where(cln => cln.ShortName == "en").Select(cln => cln.LongName).SingleOrDefault(),
-                    ca.Address.CountryAlpha2Code),
-                ca.CompanyIdentifiers.Select(ci => new UniqueIdData(ci.UniqueIdentifier!.Label, ci.Value))))
+                    new LegalAddress(
+                        ca.Address!.CountryAlpha2Code,
+                        string.Format("{0}-{1}", ca.Address!.CountryAlpha2Code, ca.Address!.Region),
```
Phil91, line 413: for my liking I'd prefer to select the fields by themselves and not do a join. I'd create a transfer object and than do the formatting in the business logic. @ntruchsess what's your opinion on that?

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
```diff
@@ -44,37 +43,14 @@ public class ClearinghouseBusinessLogic(
...
-            var walletData = await custodianBusinessLogic.GetWalletByBpnAsync(context.ApplicationId, cancellationToken);
-            if (walletData == null || string.IsNullOrEmpty(walletData.Did))
-            {
-                throw new ConflictException($"Decentralized Identifier for application {context.ApplicationId} is not set");
-            }
-
-            companyDid = walletData.Did;
-        }
-
-        await TriggerCompanyDataPost(context.ApplicationId, companyDid, overwrite, cancellationToken).ConfigureAwait(ConfigureAwaitOptions.None);
```
Phil91, line 77: Just out of curiosity, could you explain why we don't need the did anymore?

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
```diff
@@ -124,19 +105,24 @@ public async Task ProcessEndClearinghouse(Guid applicationId, ClearinghouseRespo
...
             item =>
             {
-                item.ApplicationChecklistEntryStatusId = declined
+                item.ApplicationChecklistEntryStatusId = isInvalid
                     ? ApplicationChecklistEntryStatusId.FAILED
                     : ApplicationChecklistEntryStatusId.DONE;
-                item.Comment = data.Message;
+
+                // There is not "Message" param available in the response in case of VALID so, thats why saving ClearinghouseResponseStatus param into the Comments in case of VALID only.
+                item.Comment = isInvalid
```
Phil91, line 121: Could we write all error messages concatenated in the comment field?
and maybe we could set just a fixed string for the valid case like - "Clearinghouse validation successfully executed" or something like that

Phil91 (changes requested):

Phil91:
@tfjanjua you didn't checked the `I have successfully tested my changes locally` was that on purpose? Is there a way to test if the changes are working with the new api before merging it?

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
tfjanjua: Yeah, its just not required into Tagus's new request object. I have updated the ticket with the request and response updates of Tagus release.

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
tfjanjua: Concatenating all the errors is a good idea, it will show all errors at once instead of getting an error at a time and then the other one, let me do this.

Agree, any message could be set as there is no "message" is coming from CH in case VALID.
I just wanted to avoid putting any constants so, I just set "result" field which says "VALID" (in string) in case of successful execution but please let me know, i will set the fixed string.

src/portalbackend/PortalBackend.DBAccess/Repositories/ApplicationRepository.cs
tfjanjua: done

src/portalbackend/PortalBackend.DBAccess/Repositories/ApplicationRepository.cs
tfjanjua: Well, only `CountryAlpha2Code` and `Region` is mandatory and rest are optional but I will add a check in ClearinghouseBusinessLogic.cs -> `TriggerCompanyDataPost` method.

```
if (data.LegalEntity.Address is null)
        {
            throw new ConflictException($"Address does not exist.");
        }
```

tfjanjua:
> @tfjanjua you didn't checked the `I have successfully tested my changes locally` was that on purpose? Is there a way to test if the changes are working with the new api before merging it?

Hi @Phil91 , thanks for your feedback
Actually I already have tested changes on my local but clearing house cannot access my local as clearing house calls callback url to tell us the validation results but I wanted to test the callback url call also as changes involved in callback url as well so, we have deployed this change into one of our envs. so I could test if clearing house is also able to call callback url and we are also receiving the results and processing it also ... then I will mark this checkbox  `I have successfully tested my changes locally`.

src/portalbackend/PortalBackend.DBAccess/Repositories/ApplicationRepository.cs
tfjanjua: yeah that would also be fine.
From my perspective, its fine to format params of an entity at repository level as long as I am not adding something new and just using entity param's values.

But lets listen from @ntruchsess

src/portalbackend/PortalBackend.DBAccess/Repositories/ApplicationRepository.cs
ntruchsess: Region is also nullable in the DB and needs to be checked if a value of e.g. 'DE-null' is not an acceptable value for LegalAddress.Region

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
```diff
@@ -124,19 +110,24 @@ public async Task ProcessEndClearinghouse(Guid applicationId, ClearinghouseRespo
...
+                                ? string.Join(", ", data.ValidationUnits.Where(s => s.Status != ClearinghouseResponseStatus.VALID).Select(x => x.Reason!.DetailMessage))
+                                : validData!.Status.ToString();
```
ntruchsess, line 136: you must not override a nullables type nullability without prior checking - you can do e.g. validData?.Status.ToString() ?? "" or whatever is applicable but just applying a '!' to make the compiler happy is not acceptable.
(same applies to Reason!.DetailMessage the line before)

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
```diff
@@ -99,16 +76,33 @@ private async Task TriggerCompanyDataPost(Guid applicationId, string decentraliz
...
         var transferData = new ClearinghouseTransferData(
-            data.ParticipantDetails,
-            new IdentityDetails(decentralizedIdentifier, data.UniqueIds),
-            _settings.CallbackUrl,
-            overwrite);
+            new LegalEntity(
+                data.Name,
+                new LegalAddress(
+                    data.Address.CountryAlpha2Code,
+                    string.Format("{0}-{1}", data.Address.CountryAlpha2Code, data.Address.Region),
```
ntruchsess, line 96: as Region is nullable on the db - please check whether a value like 'DE-null' is acceptable or should be adjusted accordingly.

ntruchsess (changes requested):
I did move the implicit client-side formating from the query to the business-logic adjusting the respective model-classes and unit-tests.
please my new review comments (and pull my changes before proceeding)

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
tfjanjua: At the moment we are using `IDENTIFIER` and `LEGAL_NAME` as validationModes


`IDENTIFIER` - validates whether the identifiers themselves exists, indepenedent of their relationship to the legal entity provided

`LEGAL_NAME` - validates whether the identifier is valid, and whether the name of the legal entity it is associated with matches the provided legal name


and in both cases, clearing house would not check the Address (or region) but we could make Region as required in through out the backend code and on frontend as well, as a separate activity if we decide to use `LEGAL_NAME_AND_ADDRESS` validationMode in the future because:

`LEGAL_NAME_AND_ADDRESS` - validates whether the identifier is valid, and whether the name of the legal entity, as well as the addresss it is associated with matches the provided ones.

**Note**: due to the high degree of false negatives during address matching, this method only compares the country, region and postal code.

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
tfjanjua: I see `validData?.Status.ToString() ?? ""` is an example but still I think, logically this `null` check is already there:
https://github.com/eclipse-tractusx/portal-backend/pull/1220/files/9c2f43887df651d31711cc91c71f05ba10adc5a4?diff=split&w=0#diff-a5dcf811fa3b57e7686d14046990a43050e691be31243b0729dd8d6f55e7a1deR115

and for `Reason!.DetailMessage` also.

So, when there is no VALID record so, there is definitely `Reason` object exists and otherwise not... so, apparently it might looks like satisfying the compiler.

ntruchsess: I agree on validData - I missed the isInvalid-assignment of line 123. But I guess the compiler is right - as this is passed within a lambda the compiler cannot know at which time this is being evaluated and whether the local variable isValid would eventually have changed (to bad c# doesn't support readonly local variables as e.g. java or kotlin do (using the final respective val keyword).
Reason is a different story. It is declared in the model class as nullable and it's only an assumption that Reason is contained in the ClearingHouse response when the responseStatus is not VALID. Technically this assumption is not enforced and Json-deserialization would deserialize a response that doesn't comply with your assumption with no error. You certainly rather see a log-entry like 'Response from Clearinghouse is invalid - Reason is null' (which tells you the code did expect this and the error did happen in the remote system) instead of a full Stacktrace with NullReferenceException (which tells you your own code is buggy). BTW: before net9.0 json-deserialization would not detect any mismatch of nullable declarations in model classes and actual data and one therefore must declare properties in model-classes being used for json-deserialization as nullable. As we recently did upgrade to net9 we may consider this: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/nullable-annotations

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
ntruchsess: ok - I take this as a 'yes values like 'DE-null' are acceptable ;-)

src/externalsystems/Clearinghouse.Library/BusinessLogic/ClearinghouseBusinessLogic.cs
ntruchsess: I changed the nullability-handling at Reason and take the ! at validData as justified :-)

ntruchsess (approved):

Issue #1221, opened 2025-01-06
Connectors | Connectors deletion having no SD document throws 409 error


tfjanjua:
### Current Behavior
When SD factory bit is disabled so, registered connectors does not have SD document(s). Deleting such connectors throws 409 error.

Request URL: /api/administration/Connectors/{connectorId}?deleteServiceAccount=true
Request Method: DELETE
Response Status Code: 409 Conflict

```
{
    "type": "https://tools.ietf.org/html/rfc7231#section-6.5.8",
    "title": "The resorce is in conflict with the current request.",
    "status": 409,
    "errors": {
        "Org.Eclipse.TractusX.Portal.Backend.Administration.Service": [
            "Connector status does not match a deletion scenario. Deletion declined"
        ]
    },
    "errorId": "ef369396-f9a2-4450-a523-a18042f43fd1",
    "details": [
        {
            "errorCode": "CONNECTOR_CONFLICT_DELETION_DECLINED",
            "type": "AdministrationConnectorErrors",
            "message": "Connector status does not match a deletion scenario. Deletion declined",
            "parameters": []
        }
    ]
}
```

### Expected Behavior
Connector should be able to deleted if the SD factory bit is disabled and no SD document is part of connector.

### Steps To Reproduce

1. Login to the portal
2. Navigate to "Connector Management" from Profile Icon.
3. Delete any OWNED connector that does not have SD document.

PR #1222, opened 2025-01-06
fix(connector): delete connector when sdFactory is disabled

tfjanjua:
## Description

Please include a summary of the change.

## Why

Please include an explanation of why this change is necessary as well as relevant motivation and context. List any dependencies that are required for this change.

## Issue

Ref: #1221

## Checklist

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
@@ -283,6 +283,10 @@ public class ConnectorsBusinessLogic(
             case ConnectorStatusId.PENDING:
                 await DeleteConnectorWithDocuments(connectorId, result.SelfDescriptionDocumentId.Value, result.ConnectorOfferSubscriptions, connectorsRepository);
                 break;
+            // Connector should be able to deleted if the ClearinghouseConnectDisabled bit is disabled and no SD document was part of connector.
+            case ConnectorStatusId.ACTIVE when _settings.ClearinghouseConnectDisabled:
+                await DeleteConnectorWithoutDocuments(connectorId, result.ConnectorOfferSubscriptions, connectorsRepository);
+                break;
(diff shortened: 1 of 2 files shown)


src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -283,6 +283,10 @@ public async Task DeleteConnectorAsync(Guid connectorId, bool deleteServiceAccou
             case ConnectorStatusId.PENDING:
                 await DeleteConnectorWithDocuments(connectorId, result.SelfDescriptionDocumentId.Value, result.ConnectorOfferSubscriptions, connectorsRepository);
                 break;
+            // Connector should be able to deleted if the ClearinghouseConnectDisabled bit is disabled and no SD document was part of connector.
+            case ConnectorStatusId.ACTIVE when _settings.ClearinghouseConnectDisabled:
```
Phil91, line 287: please add the result.SelfDescriptionDocumentId == null to be sure that the selfDescriptionDocumentId really isn't assigned

Phil91 (changes requested):

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
tfjanjua: Thanks for the feedback @Phil91 , I guess `ClearinghouseConnectDisabled` bit will be `false` even if the value is not assigned in settings and it will be false even if the `ClearinghouseConnectDisabled` param is not there in settings.

But I can still make `ClearinghouseConnectDisabled` as nullable and then check for the null value. Please let me know.

Phil91: made a thinking mistake. all good :)

Phil91 (changes requested):
Could you please add some unit tests to test the case of the deletion

tfjanjua:
> Could you please add some unit tests to test the case of the deletion

Thanks for pointing out @Phil91 , test case has been added

Phil91 (approved):

Issue #1175, opened 2024-11-22
OfferSetup - manage pending processes if provider callback url is removed


Phil91:
### Description
As an offer provider I'd like that the offer subscription processes are set in the correct state after I removed the provider trigger callback url 

### Acceptance Criteria

- [x] All process steps `RETRIGGER_PROVIDER` which are in `TODO` should be set to done and a new process step should be set to `AWAIT_START_AUTOSETUP` 
- [x] The callback url should be removed from the database

evegufy:
could you please also document this logic in the portal-assets?

PR #1227, opened 2025-01-08
feat(auto-setup): adjust provider callbackurl handling

Phil91:
## Description

When the provider callback url is removed the trigger provider callback steps are set to done and the await auto setup step is created

## Why

If a offer subscription process is in progress and the provider callback url is removed by the provider it might occur that the process is stuck since the next step would be to trigger the provider callback, which must not be triggered anymore 

## Issue

Refs: #1175

## Checklist

- [x] I have created and linked IP issues or requested their creation by a committer

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/SubscriptionConfigurationBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/SubscriptionConfigurationBusinessLogic.cs
@@ -18,11 +18,12 @@
  ********************************************************************************/
 
 using Org.Eclipse.TractusX.Portal.Backend.Administration.Service.ErrorHandling;
+using Org.Eclipse.TractusX.Portal.Backend.Framework.Async;
 using Org.Eclipse.TractusX.Portal.Backend.Framework.ErrorHandling;
 using Org.Eclipse.TractusX.Portal.Backend.Framework.Identity;
 using Org.Eclipse.TractusX.Portal.Backend.Framework.IO;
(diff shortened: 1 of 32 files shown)


evegufy (approved):
could you please also document this logic in the portal-assets?

src/administration/Administration.Service/BusinessLogic/SubscriptionConfigurationBusinessLogic.cs
```diff
@@ -106,9 +109,35 @@ private async Task SetOfferProviderCompanyDetailsInternalAsync(ProviderDetailDat
...
+
+    private async Task HandleOfferSetupProcesses(Guid companyId, ICompanyRepository companyRepository)
+    {
+        var processData = await companyRepository
+            .GetOfferSubscriptionProcessesForCompanyId(companyId)
+            .ToListAsync()
+            .ConfigureAwait(false);
+
+        foreach (var context in processData
+                     .Where(x => x.Process != null && x.ProcessSteps?.Any(ps => ps is
```
ntruchsess, line 132: this condition should be added to the query.

ntruchsess (changes requested):

src/administration/Administration.Service/BusinessLogic/SubscriptionConfigurationBusinessLogic.cs
Phil91: adjusted

ntruchsess (approved):

Issue #1299, opened 2025-02-26
BPDM Pool URL for Application Activation Membership Activation is Incorrect


typecastcloud:
### Current Behavior
Configured baseaddress + v6/cx-memberships is called. This ignores customizable [poolApiPath](https://github.com/eclipse-tractusx/portal/blob/ef675d0db0eb29e1e54ad3a6ecda2daec1df126e/charts/portal/values.yaml#L51).

### Expected Behavior

poolAddress/poolApiPath/cx-memberships should be called to set membership.

### Steps To Reproduce

1. Onboard company until application activation
2. Activation fails with error
 ```
call to external system https://xyz.net/v6/cx-memberships failed with statuscode 404 - Message: <html>
<head><title>404 Not Found</title></head>
<body>
<center><h1>404 Not Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

PR #1300, opened 2025-02-26
fix(BpdmService): cx-memberships call to expect custom api path

typecastcloud:
## Description

Remove hardcoded v6 path in bpdm cx-memberships api call. BusinessPartnerPoolBaseAddress should contain base api path including version.

## Why

There is a customizable poolApiPath in portal values including API version. Calls are failing if the pool has any path.

## Issue

Refs: eclipse-tractusx/portal-backend#1299
Depends on: eclipse-tractusx/portal#518

## Checklist

### DIFF
--- a/tests/externalsystems/Bpdm.Library/BpdmServiceTests.cs
+++ b/tests/externalsystems/Bpdm.Library/BpdmServiceTests.cs
@@ -675,7 +675,7 @@ public class BpdmServiceTests
         var httpMessageHandlerMock = new HttpMessageHandlerMock(HttpStatusCode.OK);
         using var httpClient = new HttpClient(httpMessageHandlerMock)
         {
-            BaseAddress = new Uri("https://business.partner.pool.base.address.com")
+            BaseAddress = new Uri("https://business.partner.pool.base.address.com/pool/v6/")
         };
         A.CallTo(() => _tokenService.GetAuthorizedClient(A<string>._, A<BpdmServiceSettings>._, A<CancellationToken>._))
             .Returns(httpClient);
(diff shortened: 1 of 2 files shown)


tests/externalsystems/Bpdm.Library/BpdmServiceTests.cs
```diff
@@ -675,7 +675,7 @@ public async Task SetCxMembership_WithValidData_DoesNotThrowException()
         var httpMessageHandlerMock = new HttpMessageHandlerMock(HttpStatusCode.OK);
         using var httpClient = new HttpClient(httpMessageHandlerMock)
         {
-            BaseAddress = new Uri("https://business.partner.pool.base.address.com")
+            BaseAddress = new Uri("https://business.partner.pool.base.address.com/pool/v6/")
```
typecastcloud, line 678: My tests fail if I don't add / at the end here. I have no idea why. @tfjanjua could you please take a look?

tfjanjua: @typecastcloud 
Validation is happening on `AbsoluteUri` and `AbsoluteUri` is a combination of `protocol` + `domain` + `path` (here `domain` means `baseAddress`) and on the other hand, normally `baseAddresses` doesn't contains the `path` (/pool/v6) and if path needs to be there then whole baseAddress must needs to end with /

For more details: https://stackoverflow.com/a/23438417

Even if you see `httpMessageHandlerMock.RequestMessage.RequestUri.AbsolutePath` or  `httpMessageHandlerMock.RequestMessage.RequestUri.PathAndQuery` (`Path` in general) so you would see `/pool/v6/cx-memberships` that the value was passed to baseAddress (`https://business.partner.pool.base.address.com/pool/v6/`) has already excluded `/pool/v6` from there.

tfjanjua (changes requested):
Change looks fine to me, please mark `I have successfully tested my changes locally` after testing the changes locally.

typecastcloud:
Successfully tested from local. Please review.

tfjanjua (approved):
@typecastcloud 
I have tested your changes on my local and its working fine
I tried with 2 BPNs

BPNL000000000065
BPNL00000000015G

and received success response from the **pool/v6/cx-memberships** API

ntruchsess (approved):

PR #1311, opened 2025-03-07
feat(onboarding): allow underscore and umlaut for commercial registration number

typecastcloud:
## Description

Add _, À-ÿ and increase maximum length for long court names in commercial registration number.

## Why

Align with portal-frontend-registration regex.

## Issue

Refs: eclipse-tractusx/portal-frontend-registration#362
Required for: eclipse-tractusx/portal-frontend-registration#364

## Checklist

### DIFF
--- a/tests/registration/Registration.Service.Tests/BusinessLogic/RegistrationBusinessLogicTest.cs
+++ b/tests/registration/Registration.Service.Tests/BusinessLogic/RegistrationBusinessLogicTest.cs
@@ -739,8 +739,14 @@ public class RegistrationBusinessLogicTest
             c.BusinessPartnerNumber == companyData.BusinessPartnerNumber);
     }
 
-    [Fact]
-    public async Task SetCompanyWithAddressAsync__WithCompanyNameChange_ModifiesCompany()
+    [Theory]
+    [InlineData(UniqueIdentifierId.VAT_ID, "DE129273398")]
+    [InlineData(UniqueIdentifierId.VIES, "DE129273398")]
+    [InlineData(UniqueIdentifierId.COMMERCIAL_REG_NUMBER, "München HRB 175450")]
+    [InlineData(UniqueIdentifierId.COMMERCIAL_REG_NUMBER, "F1103R_HRB98814")]
+    [InlineData(UniqueIdentifierId.EORI, "DE12345678912345")]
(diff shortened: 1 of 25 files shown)


tests/registration/Registration.Service.Tests/BusinessLogic/RegistrationBusinessLogicTest.cs
```diff
@@ -739,8 +739,14 @@ public async Task SetCompanyWithAddressAsync__WithExistingBpn_ModifiesCompany()
...
-    [Fact]
-    public async Task SetCompanyWithAddressAsync__WithCompanyNameChange_ModifiesCompany()
+    [Theory]
+    [InlineData(UniqueIdentifierId.VAT_ID, "DE129273398")]
+    [InlineData(UniqueIdentifierId.VIES, "DE129273398")]
+    [InlineData(UniqueIdentifierId.COMMERCIAL_REG_NUMBER, "München HRB 175450")]
+    [InlineData(UniqueIdentifierId.COMMERCIAL_REG_NUMBER, "F1103R_HRB98814")]
+    [InlineData(UniqueIdentifierId.EORI, "DE12345678912345")]
+    [InlineData(UniqueIdentifierId.LEI_CODE, "529900T8BM49AURSDO55")]
+    public async Task SetCompanyWithAddressAsync__WithCompanyNameChange_ModifiesCompany(UniqueIdentifierId uniqueIdentifierId, string identifier)
```
tfjanjua, line 749: I think this test is only checking the correct formats of the values, can we add 1-2 failure cases as well? specially the ones T-Systems has specifically recommended as AVOID

1. Just the "HRB" number: HRB 98814
While supported, usage of this format is discouraged, since the identifier is not guaranteed to be unique on the country level, only on the level of the district court. In case of multiple matches being found, the system will attempt to find the appropriate one by doing exact name matching, but this can produce false negatives. **Recommendation: avoid**

I would suggest test to avoid for both formats
1. HRB 98814
2. HRB98814

no issue if additional cases would be part of same unit test or another unit test need to be created.

tfjanjua (changes requested):

evegufy (changes requested):
https://github.com/eclipse-tractusx/portal-frontend-registration/pull/364#issuecomment-2706601341

Issue #1139, opened 2024-11-11
Create OWN connector with available Technical user only


dhiren-singh-007:
### Description

~As a user i want to see the list of only those technical user which is not connected to any connector while creating OWN connector.~
As a user i want to create an OWN connector with available(not linked to any connector or offer) technical users only.

### Acceptance Criteria

~- [ ] I should be able to fetch the list of available technical user~ After analysing this i found that existing api for technical user is returning the connector and offer info along with tech user so FE can filter out all technical users who are not linked to any offer and connector
- [ ] I should be able to send only available technical user id in the create connector api i.e  in the api `POST: /api/administration/connectors/` 
- [ ] An error should be thrown if user send the technical user id which is already linked to a connector. Api `POST: /api/administration/connectors/` . Error message : 'Techincal user id {userId} is already used by another connector or offer.' Error Type 409

~### Additional Information
It would need a new api to get the all available technical users for connectors~

SanaSalem01:
### Below cases have been tested and verified:

- [ ] I should be able to send only available technical user id in the create connector api i.e in the api POST: /api/administration/connectors/

- [ ] An error should be thrown if user send the technical user id which is already linked to a connector. Api POST: /api/administration/connectors/ . Error message : 'Techincal user id {userId} is already used by another connector or offer.' Error Type 409

<img width="753" alt="Image" src="https://github.com/user-attachments/assets/0a40f213-499a-41d5-8538-08140d1bac82" />

@dhiren-singh-007 one point, returned error is `400 Bad Request` and not 4.09, **is it intended?**

PR #1350, opened 2025-05-02
feat(connector): add validation for technical user in connector registration

dhiren-singh-007:
## Description

Added a validation for technical user in connector registration so that only available technical user can be used.

## Why

In last release database constraints has been removed and hence this api starts accepting same technical user to multiple connectors. 

## Issue

#1139 

## Checklist

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
@@ -211,14 +211,15 @@ public class ConnectorsBusinessLogic(
             return;
         }
 
-        if (!await portalRepositories.GetInstance<ITechnicalUserRepository>()
-                .CheckActiveServiceAccountExistsForCompanyAsync(technicalUserId.Value, companyId).ConfigureAwait(ConfigureAwaitOptions.None))
+        var (activeUserExists, linkedToConnectorOrOffer) = await portalRepositories.GetInstance<ITechnicalUserRepository>()
+                .CheckTechnicalUserDetailsAsync(technicalUserId.Value, companyId).ConfigureAwait(ConfigureAwaitOptions.None);
+
+        if (!activeUserExists)
         {
             throw ControllerArgumentException.Create(AdministrationConnectorErrors.CONNECTOR_ARGUMENT_TECH_USER_NOT_ACTIVE, [new ErrorParameter("technicalUserId", technicalUserId.Value.ToString()), new ErrorParameter("companyId", companyId.ToString())]);
         }
 
-        if (await portalRepositories.GetInstance<ITechnicalUserRepository>()
-        .CheckTechnicalUserLinkedToConnectorOrOfferCompanyAsync(technicalUserId.Value, companyId).ConfigureAwait(ConfigureAwaitOptions.None))
+        if (linkedToConnectorOrOffer)
         {
             throw ControllerArgumentException.Create(AdministrationConnectorErrors.CONNECTOR_ARGUMENT_TECH_USER_IN_USE, [new ErrorParameter("technicalUserId", technicalUserId.Value.ToString())]);
         }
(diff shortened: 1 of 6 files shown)


src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -212,6 +212,12 @@ private async Task ValidateTechnicalUser(Guid? technicalUserId, Guid companyId)
         {
             throw ControllerArgumentException.Create(AdministrationConnectorErrors.CONNECTOR_ARGUMENT_TECH_USER_NOT_ACTIVE, [new ErrorParameter("technicalUserId", technicalUserId.Value.ToString()), new ErrorParameter("companyId", companyId.ToString())]);
         }
+
```
Phil91, line 215: I'd suggest to align the `CheckActiveServiceAccountExistsForCompanyAsync` & `CheckTechnicalUserLinkedToConnectorOrOfferCompanyAsync` methods and return a ValueTuple with which we can work. than we only call the database once instead of two times.

Phil91 (changes requested):

evegufy:
@dhiren-singh-007 did you see Phil's review comment?

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
dhiren-singh-007: Thanks @Phil91 . Here, my view was that I want to maintain the single responsibility of the method and more modular code.Also I see that the performance impact of an additional query is likely minimal .But if you still think it's still a concern then i can change it.

Phil91: I'm a friend of combining calls to external systems whenever possible. Since `CheckActiveServiceAccountExistsForCompanyAsync` is only used in this method I'd prefer to align the two methods if that's okay with you

evegufy:
@dhiren-singh-007 will implement requested changes for beginning or next and will be part of RC2 for 25.06

src/administration/Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
dhiren-singh-007: done pls check

src/portalbackend/PortalBackend.DBAccess/Repositories/TechnicalUserRepository.cs
```diff
@@ -223,13 +223,23 @@ public void AttachAndModifyTechnicalUser(
...
+            {
+                IsActiveOrPending = tu.Identity!.UserStatusId == UserStatusId.ACTIVE || tu.Identity.UserStatusId == UserStatusId.PENDING,
+                IsLinkedToConnectorOrOffer = tu.Connector != null || tu.OfferSubscription != null
+            })
+            .SingleOrDefaultAsync();
+
+        return result != null
+            ? (result.IsActiveOrPending, result.IsLinkedToConnectorOrOffer)
+            : (false, false);
+    }
```
Phil91, line 242: you can directly return the result, if its not existing it will be set to the default which is false, false

```csharp
public async Task<(bool ActiveUserExists, bool LinkedToConnectorOrOffer)> CheckTechnicalUserDetailsAsync(Guid technicalUserId, Guid companyId) => 
    await portalDbContext.TechnicalUsers
         .Where(tu =>
             tu.Id == technicalUserId &&
             tu.Identity!.CompanyId == companyId)
         .Select(tu => new ValueTuple<bool, bool>(
             tu.Identity!.UserStatusId == UserStatusId.ACTIVE || tu.Identity.UserStatusId == UserStatusId.PENDING,
             tu.Connector != null || tu.OfferSubscription != null
         ))
         .SingleOrDefaultAsync();
```

Phil91 (changes requested):

Phil91 (approved):

Issue #1346, opened 2025-04-25
Connector | Updating connector URL throws error


tfjanjua:
### Current Behavior

Updating connector URL throws error when SD toggle is disabled and same connector was created without sd-document.

### Expected Behavior

Provider should be able to update connector URL successfully.

### Steps To Reproduce

1. Sd-Toggle is disabled (means no SD factory call will take place)
2. Login to portal app
3. Navigate to connector registration
4. Create MANAGED connector
5. Update connector URL
6. See the results.

dhiren-singh-007:
@tfjanjua : Please update this issue with issue link and also attach the PR

tfjanjua:
This issue is coming from this ticket: https://github.com/eclipse-tractusx/portal-backend/issues/549

saadanzari:
@ntruchsess to add review in PR
cc @Phil91 @evegufy

PR #1357, opened 2025-05-12
fix(connector): sd-toggle check while updating connector url

tfjanjua:
## Description

Updating connector URL throws error

## Why
Updating connector URL throws error when SD toggle is disabled and same connector was created without sd-document.
- Consistency of create, update and delete connector while getting provider company's country code.
- Get provider company's country code to get country-specific settings.
- Get country-specific settings to get correct sd-toggle to call sd-factory API.

## Issue

Ref: #1346

## Checklist

### DIFF
--- a/src/portalbackend/PortalBackend.DBAccess/Models/OfferSubscriptionWithProvider.cs
+++ b/src/portalbackend/PortalBackend.DBAccess/Models/OfferSubscriptionWithProvider.cs
@@ -1,6 +1,6 @@
 /********************************************************************************
- * Copyright (c) 2022 BMW Group AG
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023 BMW Group AG
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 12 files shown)


tfjanjua:
@Phil91 | @MaximilianHauer | @dorkabcox 

There is some sort of connection between company's self-description document and company's connector's self-description document.

When we register a company, we send company's bpn to clearinghouse via sd-factory to generate self-description document and then we receive a signed document from directly from clearinghouse and we save this document in portal db with recently registered company.

Now, if the same company tries to register its Owned connector (COMPANY_CONNECTOR):

- we assign logged in user's company as provider and host/customer
- and send logged in user's company's BPN to clearinghouse via sd-factory
- along with BPN, we use logged in user's company's SelfDescriptionDocumentId and generate a url and send it to clearinghouse via sd-factory so that sd-factory (or clearinghouse) could get company's document and would know that this coming data of connector is related to an already registered company.

Now, if the same company tries to register its Managed connector (CONNECTOR_AS_A_SERVICE):

- We assign logged in user's company as provider and subscriber user's company as host/customer
- and we send host/customer company's BPN to clearinghouse via sd-factory here
- along with BPN, we use host/customer company's SelfDescriptionDocumentId and generate a url and send it to clearinghouse via sd-factory so that sd-factory (or clearinghouse) could get company's document and would know that this coming data of connector is related to an already registered company.

However, Provider is the one who has the information of connector and Provider is the one who registers the connector for one if his offer's subscribers as host/customer so, I think this is the point where we would need to confirm if we should use provider's info (like bpn and SelfDescriptionDocumentId) instead of host/customer's in Managed Connector?

I have made create, update and delete connector methods consistent by getting bpn and SelfDescriptionDocumentId of a connector Provider.

and as I understand it, countryCode should come from the address of a company whose bpn we send to sd-factory?

and if I shouldn't link countryCode to a company's bpn, then I think we can use the connector's country code directly, regardless of whose bpn we end up sending to sd-factory?

I am actually not aware of business logic behind the Connector's countryCode.

The reason I want to understand this is because I have enabled/disabled the sd-factory service by getting country code of Provider's address, otherwise the Connector itself has countryCode, then provider and then host/customer as well.

dorkabcox:
hello @ntruchsess and/or @Phil91 could you please kindly review this PR? we would love to move ahead with our Io release. Thank you! :)

MaximilianHauer:
@tfjanjua thank you for the explanation i understand where you want to go with your ideas but i have to tell you that i was not involved in this implementation myself. i will try to see if i can find someone who was part and could provide a proper answer.

src/portalbackend/PortalBackend.DBAccess/Models/OfferSubscriptionWithProvider.cs
```diff
@@ -0,0 +1,45 @@
+/********************************************************************************
+ * Copyright (c) 2023 BMW Group AG
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
```
evegufy, line 3:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

evegufy (changes requested):
please rebase to release branch if it should be part of 2.5.0/25.06

Phil91 (changes requested):
One minor change request. We should wait with the merge until the open questions about the logic are clarified

tfjanjua:
> @tfjanjua thank you for the explanation i understand where you want to go with your ideas but i have to tell you that i was not involved in this implementation myself. i will try to see if i can find someone who was part and could provide a proper answer.

hi @MaximilianHauer , please let me know if you have any information, we would need to push this fix in 25.06.

evegufy:
@tfjanjua please resolve conflict.
@ntruchsess could you please provide your point of view to this comment https://github.com/eclipse-tractusx/portal-backend/pull/1357#issuecomment-2871955762?

typecastcloud:
@tfjanjua kept as known known for 25.06. Branch changed to main after exclusion.

cc: @evegufy

ntruchsess (approved):

ntruchsess (approved):

typecastcloud (approved):

Issue #1375, opened 2025-05-30
Auto-Setup Not Triggered on App Subscription with new auto setup configuration (Auto Setup URL, Auth URL, Client ID, and Client Secret)


sachinargade123:
### Current Behavior

As the app provider, In Catenax INT environment, we have configured the new auto-setup settings, including the Auto Setup URL, Auth URL, Client ID, and Client Secret.

Attempt 1: Using Existing Approved App – "Connect & Integrate"

We attempted to unsubscribe from the current subscription and then resubscribe to the "Connect & Integrate" app via the marketplace. However, no request was received from the portal backend during this process.

As a result, despite the auto-setup being correctly configured, the subscription remains in the PENDING state processStepTypeId: RETRIGGER_PROVIDER.

Below are the subscription details:
> 
    {
    "offerId": "ce71a1c5-f622-49d5-9e3b-75b1cc5944e3",
    "offerName": "Connect & Integrate",
    "companySubscriptionStatuses": [
        {
                   "companyId": "d3f4e669-f164-47bd-9062-1d17b9ad7406",
                    "companyName": "Mehrans Eisdiele",
                    "subscriptionId": "192a59c8-b02a-4922-8511-0dce8ed70a69",
                    "offerSubscriptionStatus": "PENDING",
                    "country": "DE",
                    "bpnNumber": "BPNL000000000ISY",
                    "email": "Sachin.Argade@t-systems.com",
                    "technicalUser": true,
                    "processStepTypeId": "RETRIGGER_PROVIDER"
      }]
> 

Attempt 2: Using a Newly Released App – "test23"

We also released a new app named "test23" in the app marketplace (not yet active). We attempted to subscribe to it, but this subscription also remains in the PENDING state (processStepTypeId: RETRIGGER_PROVIDER), and again, no request was received by the app backend.
>

    {
            "offerId": "3b9aaa24-b1d8-475f-b927-39a88033a1b6",
            "offerName": "test23",
            "companySubscriptionStatuses": [
                {
                    "companyId": "d3f4e669-f164-47bd-9062-1d17b9ad7406",
                    "companyName": "Mehrans Eisdiele",
                    "subscriptionId": "369597c7-f311-46e1-bbc6-a6926917cb9f",
                    "offerSubscriptionStatus": "PENDING",
                    "country": "DE",
                    "bpnNumber": "BPNL000000000ISY",
                    "email": "Dheeraj.Bhargava@t-systems.com",
[...]

PR #1376, opened 2025-06-02
fix(auto-setup): fixed grant type

dhiren-singh-007:
## Description

Added grant type in token generation in auto setup.

## Why

Token generation was failing as most of the IDP requires grant type

## Issue

#1375 

## Checklist


typecastcloud (commented):
Hi @dhiren-singh-007, this requires you to create a new nuget package with this change.
See src/framework/README.md

Other clients also set scope to "openid". I'm not sure how defaults work for this.

dhiren-singh-007:
> Hi @dhiren-singh-007, this requires you to create a new nuget package with this change. See src/framework/README.md
> 
> Other clients also set scope to "openid". I'm not sure how defaults work for this.

I think it's only required when we change in Framework project inside framework folder.
"openid" is optional.

Issue #1259, opened 2025-02-04
Update all apis to return company short name


dhiren-singh-007:
### Description
As a frontend user, I want to get the legal name and short legal name from the following APIs so that it would be flexible to show the names according to need.
Some companies can have very long names, so it would be best to show a short name in the front end.

### Acceptance Criteria
Apps Api
- [ ] `api/apps/active` should return provider's short name, i.e providerShortName
- [ ] `api/apps/business` should return provider's short name, i.e providerShortName
- [ ] `api/apps/{appId}` should return provider's short name, i.e providerShortName
- [ ] `api/apps/provided` should return provider's short name, i.e providerShortName
- [ ] `api/apps/subscribed/subscription-status` should return provider's short name, i.e providerShortName
- [ ] `api/apps/{appId}/subscription/{subscriptionId}/provider` should return customer's short name, i.e customerShortName
- [ ] `api/apps/{appId}/subscription/{subscriptionId}/subscriber` should return provider's short name, i.e providerShortName
- [ ] `api/apps/subscribed/activesubscriptions` should return provider's short name, i.e providerShortName
- [ ] `api/apps/appreleaseprocess/{appId}/appStatus` should return provider's short name, i.e providerShortName
- [ ] `api/apps/appreleaseprocess/inReview` should return provider's short name, i.e providerShortName
- [ ] `api/apps/appreleaseprocess/inReview/{appId}` should return provider's short name, i.e providerShortName

Services Api
- [ ] `api/services/active` should return provider's short name, i.e providerShortName
- [ ] `api/services/{serviceId}` should return provider's short name, i.e providerShortName
- [ ] `api/services/serviceConsent/{serviceConsentId}` should return company's short name, i.e companyShortName
- [ ] `api/services/provided` should return provider's short name, i.e providerShortName
- [ ] `api/services/{serviceId}/subscription/{subscriptionId}/provider` should return customer's short name, i.e customerShortName
- [ ] `api/services/{serviceId}/subscription/{subscriptionId}/subscriber` should return provider's short name, i.e providerShortName
- [ ] `api/services/subscribed/subscription-status` should return provider's short name, i.e providerShortName
- [ ] `api/services/servicereleaseprocess/{serviceId}/appStatus` should return provider's short name, i.e providerShortName
- [ ] `api/services/servicereleaseprocess/inReview` should return provider's short name, i.e providerShortName
[...]

PR #1387, opened 2025-06-06
feat: update registration administration apis to return company short name

JordanGerada3:
## Description

Added host/provider/company short name attributes to service API.

## Why

Some companies can have very long names, to improve usability/readability in the UI, it would be best to allow for company short name to be displayed.

## Issue

#1259 

## Checklist

### DIFF
--- a/src/portalbackend/PortalBackend.DBAccess/Models/ConnectorData.cs
+++ b/src/portalbackend/PortalBackend.DBAccess/Models/ConnectorData.cs
@@ -20,6 +20,7 @@
 
 using Org.Eclipse.TractusX.Portal.Backend.PortalBackend.PortalEntities.Enums;
 using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
 
 namespace Org.Eclipse.TractusX.Portal.Backend.PortalBackend.DBAccess.Models;
 
@@ -60,6 +61,7 @@ public record ManagedConnectorData(
     ConnectorTypeId Type,
     ConnectorStatusId Status,
     string? ProviderCompanyName,
+    [property: JsonPropertyName("providerShortName")] string Shortname,
     Guid? SelfDescriptionDocumentId,
     TechnicalUserData? TechnicalUser,
     string ConnectorUrl);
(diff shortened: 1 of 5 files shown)


src/portalbackend/PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
```diff
@@ -230,8 +233,9 @@ public void DeleteConnectorAssignedSubscriptions(Guid connectorId, IEnumerable<G
                 con.Id,
                 con.TypeId,
                 con.Name,
-                con.HostId ?? con.ProviderId,
                 con.HostId != null ? con.Host!.Name : con.Provider!.Name,
+                con.Host!.Shortname,
```
tfjanjua, line 237: I think here we should put checks same as link# 236 (`con.HostId != null ? con.Host!.Name : con.Provider!.Name`)

we can put check on line# 237 like: `con.HostId != null ? con.Host!.ShortName : con.Provider!.ShortName`

src/portalbackend/PortalBackend.DBAccess/Models/ConnectorData.cs
```diff
@@ -43,26 +44,23 @@ string ConnectorUrl
...
 public record ManagedConnectorData(
     string Name,
-    [StringLength(2, MinimumLength = 2)]
-    string Location,
+    [StringLength(2, MinimumLength = 2)] string Location,
     Guid Id,
     ConnectorTypeId Type,
     ConnectorStatusId Status,
     string? ProviderCompanyName,
+    [property: JsonPropertyName("hostCompanyShortName")] string? Shortname,
```
tfjanjua: here we have `ProviderCompanyName` on line# 58 but you have mention `hostCompanyShortName` for shortname so, is it fine? like shouldn't we get Providers' shortname instead of host's?

I see in the following link, you have set Provider's shortName so I think this should be `providerCompanyShortName`
https://github.com/eclipse-tractusx/portal-backend/pull/1387/files#diff-95871bb6a566d615990a1a2d43f14dc73f97787e700f0be901b33ee5df1573cfR88

tfjanjua (changes requested):

src/portalbackend/PortalBackend.DBAccess/Models/ConnectorData.cs
JordanGerada3: @tfjanjua to remain consistent with the ProviderCompanyName I have switched the naming to user provider. Thank you for the feedback.

src/portalbackend/PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
JordanGerada3: Added.

Issue #1458, opened 2025-10-20
Update connector URL | 409 updating connector url with technical user token


JordanGerada3:
When using the token generated by clientId and secret of technical user created with Connector Management profile, user tries to update the connector URL 409 error displays with following error details:

`errorId": "********",
    "details": [
        {
            "errorCode": "CONNECTOR_CONFLICT_SET_BPN",
            "type": "AdministrationConnectorErrors",
            "message": "The bpn of company {companyId} must be set",
            "parameters": [
                {
                    "name": "companyId",
                    "value": "********"
                }
            ]
        }
    ]
}`

### Current Behavior
409 error when updating the connector URL with the token issued via the associated technical user for that connector. 

### Expected Behavior

URL can be updated via the UI and via the associated technical user generated token.

### Steps To Reproduce

- Login to portal UI
- Create a technical user 'Internal technical user profile'
- Select role 'Connector Management'
- Register a connector and configure with the created technical user
- Create a token with the technical user created
- In Swagger use API: PUT /api/administration/connectors/{connectorId}/connectorUrl

**NOTE**

- If the token to make the call is the token generated via portal UI, API request is successful
- If the token used to make the call is generated via the technical user, API request throws an error

PR #1459, opened 2025-10-20
fix: get BPN from company id update connector url

JordanGerada3:
## Description

Refactor BPN fetching function to get BPN based on company id and not user id.

## Why

To update a connector url, a user should be able to update it via Portal UI or an external API call.

In the case of updating the url via the portal UI, this functions correctly. However, when updating the URL externally via a technical user(Connector Management) access token, a 409 error is thrown "The bpn of company {companyId} must be set".

This issue arrises as the function that gets the company BPN relies on the user id of the current logged in user. When accessing the API via the technical user(Connector Management), the mapping from user -> company is lost and therefore a BPN cannot be retrieved and set.

## Issue

[1458](https://github.com/eclipse-tractusx/portal-backend/issues/1458)

## Checklist

### DIFF
--- a/tests/portalbackend/PortalBackend.DBAccess.Tests/UserRepositoryTests.cs
+++ b/tests/portalbackend/PortalBackend.DBAccess.Tests/UserRepositoryTests.cs
@@ -230,9 +230,7 @@ public class UserRepositoryTests : IAssemblyFixture<TestDbFixture>
     {
         // Arrange
         var (sut, context) = await CreateSutWithContext();
-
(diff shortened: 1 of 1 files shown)


tfjanjua:
@JordanGerada3 I think there are some tests failing so please have a look at them, and I would have questions: which tech user here we are taking about and what permissions that user may have and if that user could call some other APIs as well and company's user id is required there as well? so should be replace company's userId with companyId to those places as well? or should we see if the mapping of user -> company can be done?

There is a possibility to have the same issue with other tech users as well and then this could be a small change but at so many places but lets work on this one tech user first.

Issue #1473, opened 2025-11-27
Too much details are being displayed in generic error message owncompany/identityproviders (PUT)


JordanGerada3:
When making a request via the owncompany/identityproviders (PUT) endpoint too much information is being reviewed which may pose a security risk.

### Current Behavior

The following information is returned when throwing a generic error: 
- Internal component names
- URLs
- Keycloak import-config endpoint

### Expected Behavior

- Only specific information about the failure to assist in debugging. 
- No sensitive or potentially compromising information to be displayed. 

### Steps To Reproduce

-  Login to the Portal
- Navigate to the IDP Management page
- Click on the Add Identity Provider button
- Enter necessary details till you reach the Metadata URL input box. Enter any random URL. Click on the Save Metadata button and intercept the request.

PR #1474, opened 2025-11-28
fix: improve error handling and remove system data from error message idp managment

JordanGerada3:
## Description

Improved error handling and less revealing error response messages. 

## Why

The application consistently returns overly verbose error responses that expose internal implementation details whenever an external request or backend process fails. These responses include internal namespaces, library identifiers, upstream service URLs, and backend component references that should not be visible to end users.
Such information disclosure reveals aspects of the application’s architecture, technologies, and integration points, which could help an attacker map the internal environment and craft more targeted reconnaissance or exploitation attempts.

## Issue

https://github.com/eclipse-tractusx/portal-backend/issues/1473

## Checklist

### DIFF
--- a/tests/framework/Framework.ErrorHandling.Web.Tests/Framework.ErrorHandling.Web.Tests.csproj
+++ b/tests/framework/Framework.ErrorHandling.Web.Tests/Framework.ErrorHandling.Web.Tests.csproj
@@ -45,7 +45,7 @@
 
     <ItemGroup>
       <ProjectReference Include="..\..\..\src\framework\Framework.ErrorHandling.Web\Framework.ErrorHandling.Web.csproj" />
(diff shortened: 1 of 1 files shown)


src/provisioning/Provisioning.Library/Extensions/IdentityProviderManager.cs
```diff
@@ -81,6 +81,7 @@ private async ValueTask<IdentityProvider> SetIdentityProviderMetadataFromUrlAsyn
         {
             throw new ServiceException("failed to import identityprovider metadata", HttpStatusCode.NotFound);
         }
+
```
tfjanjua: I do not see any change in this file but just one line space, can we remove this?

tfjanjua (changes requested):

